PR #2, opened 2017-02-21
Implement ToSql for Postgres

shssoichiro:
I found this library because I need a type that can represent a numeric/decimal Postgres type, but discovered that the `postgres` feature wasn't fully implemented and didn't compile. This PR fixes the compilation issues and adds a `ToSql` implementation. The implementation may not be the most efficient, but it should be sufficiently fast for day-to-day use.


paupino:
Cool, thanks for doing this. Will publish a new version shortly.

PR #3, opened 2017-02-21
Bump serde to 0.9

shssoichiro:
This PR bumps the serde version to 0.9, which compiles on stable as of Rust 1.15. This removes the need for the `syntex` dependency, which should significantly reduce compile times.


paupino:
Eventually I'll move serde into it's own feature. All in all, looks good.

PR #4, opened 2017-02-24
Bump dependencies, most notably postgres to 0.14

shssoichiro:
Note that postgres 0.14 is not backwards compatible with 0.13, because definitions of `from_sql` and `to_sql` have changed.


paupino:
Looks good; thanks for that.

PR #12, opened 2017-06-09
Replace ToDecimal with From/Into

### LINKED ISSUE #8: Replace ToDecimal with Into and From?
I was having a quick look through your code, and I see you have a new trait called `ToDecimal` which simply converts integer types into a decimal.  I was curious as to why you didn't opt for the [`Into`](https://doc.rust-lang.org/std/convert/trait.Into.html) and [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) traits from the standard library which do exactly that?

Would you like a pull request to remove the `ToDecimal` trait and replace it with the standard library equivalent?

JP-Ellis:
From and Into are traits from the Rust standard library intended to be used
precisely when types need to be converted into one another (without failing).

Closes #8.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -370,32 +370,27 @@ impl Decimal {
     }
 }
 
-pub trait ToDecimal {
-    /// Converts the value of `self` to a `Decimal`.
-    fn to_decimal(&self) -> Option<Decimal>;
-}
-
-macro_rules! impl_to_decimal {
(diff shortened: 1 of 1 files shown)


paupino:
Looks good. Technically this breaks the API so I _should_ really do a major version change however for now will just bump the minor version. In the future I think we'll need to look at adding the `deprecated` attribute.

PR #22, opened 2017-06-10
Run rustfmt on all the code

JP-Ellis:
I'm not sure whether this can (and should) be automated with Travis or something?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,8 @@
 extern crate num;
 extern crate serde;
 extern crate serde_json;
-#[macro_use] extern crate lazy_static;
+#[macro_use]
+extern crate lazy_static;
 
 #[macro_use]
 #[cfg(feature = "postgres")]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -183,7 +182,8 @@ impl Decimal {
 
         let old_scale = self.scale();
 
-        if dp < old_scale && dp < 20 { // Technically, it's 28...
+        if dp < old_scale && dp < 20 {
+            // Technically, it's 28...
```
paupino: This comment is not great, but loses context here. In fact, I think it's probably a bug which needs another issue created (e.g. why can't I round more than 20 dp?).

src/postgres.rs
```diff
@@ -170,27 +165,42 @@ impl ToSql for Decimal {
...
         } else {
             whole_portion.len() as i16 - 1
         };
-        let all_groups = whole_portion.into_iter().chain(decimal_portion.into_iter())
-            .skip_while(|ref x| *x == "0000").collect::<Vec<String>>();
+        let all_groups = whole_portion
+            .into_iter()
+            .chain(decimal_portion.into_iter())
+            .skip_while(|ref x| *x == "0000")
+            .collect::<Vec<String>>();
```
paupino: These are all much cleaner!

paupino:
Will merge however we should address that one comment via an issue (it's limited thinking from what I can determine right now - if it's not, then the comment isn't doing justice).

src/lib.rs
JP-Ellis: The comment doesn't have much context in the first place.  I don't think the change from rustfmt made things any better or worse.

paupino: Created #23 to address

PR #27, opened 2017-06-10
Library error

JP-Ellis:
Making `&'static str` the public error type means that it cannot be changed in the future as that would break the interface.  Using a library-specific Error type allows for non-breaking changes in the future.

This will need to be merged *after* #26 (though I'm not sure whether it'll work...)


src/error.rs
```diff
@@ -0,0 +1,32 @@
+use std::error;
+use std::fmt;
+
+/// Error type for the library.
+#[derive(Clone,Debug)]
+pub struct Error {
```
paupino: I was thinking of using `error_chain` at some stage in the future, but for now this is definitely a move in the right direction.

PR #30, opened 2017-06-21
Replace ToString with fmt::Display

JP-Ellis:
Should be merged after #29.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -3,6 +3,7 @@ use num::{BigInt, BigUint, FromPrimitive, Integer, One, ToPrimitive, Zero};
 use num::bigint::Sign::{Minus, Plus};
 use std::cmp::*;
 use std::cmp::Ordering::Equal;
+use std::fmt;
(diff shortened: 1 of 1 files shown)


src/decimal.rs
```diff
@@ -664,6 +624,45 @@ impl ToPrimitive for Decimal {
...
+                rep.insert(len - scale as usize, '.');
+            }
+        }
+
+        // Last set the negation
+        if self.is_negative() {
+            rep.insert(0, '-');
+        }
+
+        f.pad(&rep)
```
paupino: Is there a reason you're using `pad` vs `write_str` here?

JP-Ellis: If I'm not mistaken, this `pad` follows certain format directives unlike `write_str` which simply writes the string.  For example, `format!("{:<30}", d)` should right-align the decimal number in a column that is 30 characters wide and `write_str` would not do that.

paupino: :+1:

PR #32, opened 2017-08-20
Remove static reference for POWERS_10

nvzqz:
The values remain static but you don't have this (in my opinion) weird layer of indirection applied. This is essentially a cosmetic change that doesn't affect functionality.


paupino:
Looks good - thanks for the contribution!

PR #34, opened 2017-08-21
Adapt to postgres 0.15

jdroenner:
(and [removed the Cargo.lock file](http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries))


paupino:
Thanks; looks good.

PR #37, opened 2017-08-27
Interoperation with f64

michaelmattig:
In order to calculate e.g. weighted averages of Decimal values, it is necessary to convert Decimals into floating point representation and accept the loss in precision.

Furthermore, it is a common use case to get financial data in JSON format. In order to automatically deserialize Decimals from JSON via serde I propose a visitor_f64.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -636,6 +636,26 @@ impl FromPrimitive for Decimal {
 }
 
 impl ToPrimitive for Decimal {
+
+    fn to_f64(&self) -> Option<f64> {
+        if self.scale() == 0 {
+            let bytes = self.unsigned_bytes_le();
+            let sign;
+            if self.is_negative() {
+                sign = Minus;
+            } else {
+                sign = Plus;
+            }
+            
+            BigInt::from_bytes_le(sign, &bytes[..]).to_f64()
+        } else {
+            match self.to_string().parse::<f64>() {
+                Ok(s) => Some(s),
+                Err(_) => None
+            }
+        }
+    }
+
     fn to_i64(&self) -> Option<i64> {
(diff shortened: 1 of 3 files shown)


src/serde_types.rs
```diff
@@ -18,6 +19,13 @@ struct DecimalVisitor;
 impl<'de> serde::de::Visitor<'de> for DecimalVisitor {
     type Value = Decimal;
 
+    fn visit_f64<E>(self, value: f64) -> Result<Decimal, E> {
+        match Decimal::from_str(&value.to_string()) {
```
paupino: I've started promoting errors in the serde library recently. Are we able to get this changed to:

`Decimal::from_str(&value.to_string()).map_err(|_| E::invalid_value(Unexpected::Float(value), &self))`

src/decimal.rs
```diff
@@ -636,6 +636,26 @@ impl FromPrimitive for Decimal {
...
+            let sign;
+            if self.is_negative() {
+                sign = Minus;
+            } else {
+                sign = Plus;
+            }
+            
+            BigInt::from_bytes_le(sign, &bytes[..]).to_f64()
+        } else {
+            match self.to_string().parse::<f64>() {
```
paupino: I wonder if I need to look at the overflow cases in further detail going forward? Just thinking that at the moment it'd return `None` which may result in unintended functionality. This code as it stands is consistent with the rest of the library so should be ok for now.

michaelmattig: What overflows are you thinking of? Decimal::max_value() does not overflow f64. It does result in a loss of precision, but I think that's OK.

paupino:
👍

PR #48, opened 2017-10-05
Fixes #35

### LINKED ISSUE #35: Implement format instructions
The following code:
```
let a = Decimal::new(100, 2);
let b = Decimal::new(300, 2);
let c = a / b;
println!("{}", c);
println!("{:.2}", c);
```
produces the output:
```
0.3333333333333333333333333333
0.
```

This is because the implementation converts to a string first and applies the `pad` method. Decimal should implement further format instructions to support producing 2dp without requiring rounding first up.

jekirl:
Adds support for formatting instructions as per #35

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -699,6 +699,10 @@ impl fmt::Display for Decimal {
         let mut rep = uint.to_string();
         let len = rep.len();
 
+        if let Some(n_dp) = f.precision() {
+            rep.truncate(len - (scale as usize) + n_dp)
+        }
+
         // Inject the decimal point
         if scale > 0 {
             // Must be a low fractional
@@ -717,12 +721,7 @@ impl fmt::Display for Decimal {
             }
         }
 
-        // Last set the negation
-        if self.is_negative() {
-            rep.insert(0, '-');
-        }
-
-        f.pad(&rep)
+        f.pad_integral(self.is_positive(), "", &rep)
     }
 }
 
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -646,7 +646,7 @@ impl ToPrimitive for Decimal {
             } else {
                 sign = Plus;
             }
-            
+
             BigInt::from_bytes_le(sign, &bytes[..]).to_f64()
         } else {
             match self.to_string().parse::<f64>() {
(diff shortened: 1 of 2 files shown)


tests/decimal_tests.rs
```diff
@@ -85,6 +85,31 @@ fn it_can_serialize_deserialize() {
     assert_eq!("12.3456789", b.to_string());
 }
 
+// Formatting
+
+#[test]
+fn it_formats() {
+    let a = Decimal::from_str("233.323223").unwrap();
+    assert_eq!(format!("{}", a), "233.323223");
+    assert_eq!(format!("{:.9}", a), "233.323223");
```
paupino: Shouldn't this have 9dp padded? i.e. `233.323223000`?

tests/decimal_tests.rs
```diff
@@ -85,6 +85,31 @@ fn it_can_serialize_deserialize() {
...
+    assert_eq!(format!("{:.0}", a), "233.");
```
paupino: Should be `233` with no trailing `.`

tests/decimal_tests.rs
```diff
@@ -85,6 +85,31 @@ fn it_can_serialize_deserialize() {
...
+    assert_eq!(format!("{:.2}", a), "233.32");
+    assert_eq!(format!("{:010.2}", a), "0000233.32");
+    assert_eq!(format!("{:0<10.2}", a), "233.320000");
+}
+#[test]
+fn it_formats_neg() {
+    let a = Decimal::from_str("-233.323223").unwrap();
+    assert_eq!(format!("{}", a), "-233.323223");
+    assert_eq!(format!("{:.9}", a), "-233.323223");
```
paupino: As above, should be `-233.323223000`

tests/decimal_tests.rs
```diff
@@ -85,6 +85,31 @@ fn it_can_serialize_deserialize() {
...
+    assert_eq!(format!("{:.0}", a), "-233.");
```
paupino: As above, no trailing `.`

src/decimal.rs
```diff
@@ -717,12 +721,7 @@ impl fmt::Display for Decimal {
             }
         }
 
-        // Last set the negation
-        if self.is_negative() {
-            rep.insert(0, '-');
-        }
-
-        f.pad(&rep)
+        f.pad_integral(self.is_positive(), "", &rep)
```
paupino: 👍

jekirl:
Fixed the case where `precision > scale`. I had intentionally not add padded 0s since the behavior for `precision > MAX_PRECISION` could be a bit unexpected (as perceived increase in decimal precision).  Also added more tests and a fix for the corner case of ending up with an empty string after truncating.

paupino:
Looks good! I'm think your concern for increasing precision is warranted. I think perhaps in the future `error-chain` could be utilized and I replace some of the panic's with `Result`. The initial reason for `panic` doesn't hold true now with `?` and `map_err` etc.

Issue #14, opened 2017-06-10
Create Benchmark tests for common operations


paupino:
There are a number of functions in the library that can be optimized, however before optimization it'd be nice to be able to measure performance (so that performance gains are measured).

jekirl:
I can take care of this. Would probably also be a good idea to have benchmarks comparing to the decimate and decimal crates.

PR #50, opened 2017-10-06
add basic benchmark against d128, closing #14

jekirl:
Basic benchmarks comparing 4 primary ops to the d128 crate that is based on FFI to decNumber


paupino:
Thanks for this! As expected, division is performing badly (I used long form) however add/mul/sub could definitely get some simple tweaks to speed it up. At least now there is something to compare it to!

PR #56, opened 2017-10-22
parse: don't panic on empty strings

### LINKED ISSUE #55: panic when parsing an empty string
`let x: Result<Decimal, Error> = "".parse()` panics. It looks like the implementation assumes the string is non-empty.

BurntSushi:
Fixes #55


paupino:
Thanks for the contribution! Looks good.

PR #64, opened 2017-11-10
Speed up same scale comparison

arkaitzj:
This change allows for fast cmp as long as you keep your decimals to the same scale.

115ns per comparison vs 22ns per comparison in my laptop.


paupino:
Good timing; I've got a PR coming soon which should further assist with different scales also.

Issue #52, opened 2017-10-10
Optimize div operation


paupino:
The current implementation of `div` focused on correctness over performance therefore can use some optimization.

Current benchmarks on 2.3GHz i7, 16GB DDR3, macOS High Sierra:
```
test bench_decimal_div ... bench:  47,344,736 ns/iter (+/- 6,536,510)
```

paupino:
Will try to reduce dependency on BigInt's etc across `rescale` and `round_dp` during this.

paupino:
New numbers:
```
test bench_decimal_div ... bench:   3,961,951 ns/iter (+/- 328,683)
```

paupino:
The changes from #95 are subtle on a like for like comparison, but still regularly faster:
```
test bench_decimal_div ... bench:   3,349,243 ns/iter (+/- 131,503)
```

paupino:
I think it's worthwhile reworking a lot of this logic to better align with long division. This will improve the maintainability of the code plus also open up some quick performance improvements.

PR #65, opened 2017-11-10
First attempt at division optimization

paupino:
Future requests will also rid of BigInt/BigUint temp variables.

Ideally I want to get this under 1,000,000 ns/iter.

Old:
```
test bench_decimal_div ... bench:  47,344,736 ns/iter (+/- 6,536,510)
```

New:
```
test bench_decimal_div ... bench:   3,961,951 ns/iter (+/- 328,683)
```


paupino:
Relates to #52

Issue #11, opened 2017-06-08
Stabalize the shorthand Decimal notation


paupino:
Creating a new decimal object is rather verbose. It'd be nice to have a shorthand option to define a decimal such as: `12.4M`. This could possibly be implemented via a `build.rs` file.

paupino:
Using `build.rs` may be a bit pre-emptive since we need to write files for this to work. This could work with an external library using `proc_macro_derive`.

paupino:
This may require

paupino:
This may require some sort of plugin like [`proc_macro_plugin`](http://manishearth.github.io/rust-internals-docs/proc_macro_plugin/index.html).

paupino:
A nightly only implementation has been created which relies on stablilization of [#38356](https://github.com/rust-lang/rust/issues/38356).

There may be an opportunity to leverage [procedural_masquerade](https://docs.rs/procedural-masquerade/0.1.5/procedural_masquerade/) to ensure this works on stable also.

gabisurita:
```
```

paupino:
One small change you could also make to avoid the additional `to_string` is to use `stringify`:

```rust
macro_rules! dec {
    ($e:expr) => {
        {
            use std::str::FromStr;
            Decimal::from_str(stringify!($e)).unwrap()
        }
    };
}
```

Hopefully this feature will be stabalizing soon with Rust 2018!

PR #70, opened 2017-11-11
Introduces Procedural Macro for Decimal type

paupino:
This adds in an extra crate to assist in a procedural macro definition for the Decimal type. This only works on nightly at the moment as it requires stabilization of [#38356](https://github.com/rust-lang/rust/issues/38356)


paupino:
This is version 1 which requires nightly. Before merge, I will try a secondary version that targets stable also.

paupino:
Addresses #11 somewhat.

Issue #73, opened 2017-11-22
Extend test suites


paupino:
Extend the test suites so that the tests are more dynamic. e.g. loop through an array of examples + results asserting along the way.

As part of this I want to also implicitly test transitive operations as appropriate. e.g. a + b = b + a, a * b = b * a.

PR #75, opened 2017-11-22
Added simple test suites for common ops with transitive alternatives where appropriate

paupino:
As a result, moved ordering tests to integration test suite since it isn't using private properties.
This looks like tests are being removed however it's actually adding far more tests than previously due to:

1. Transitive tests for mul and add
2. Additional eq and cmp tests
3. Plus one for parsing underflow

All in all, it provides a method to easily extend the tests for any edge cases found.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1136,6 +1136,11 @@ impl FromStr for Decimal {
             0
         };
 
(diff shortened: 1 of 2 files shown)


paupino:
Fixes #73

PR #88, opened 2017-12-28
add test case for failing addition

rkusa:
Hi,

here is a test case for an addition that returns a wrong result.

```
---- it_adds_decimals stdout ----
	thread 'it_adds_decimals' panicked at 'assertion failed: `(left == right)`
  left: `"12.45378151260504201681"`,
 right: `"4.5309652611786082574523276562"`: 11.815126050420168067226890757 + 0.6386554621848739495798319328', tests/decimal_tests.rs:168:8
```

The result of `12.45378151260504201681`, which I've added to the test case, is maybe not the exact expected one, but the current result of `4.5309652611786082574523276562` seems to be wrong for sure.

Thanks


paupino:
Thank you. Rather than merge this I'm going to recreate in a PR which fixes it to avoid breaking the master branch.

rkusa:
Sure! The PR wasn't meant to be merged, so please feel free to close the PR, thanks!

paupino:
Just as an early diagnosis - this is due to the re-scaling logic silently overflowing. Essentially, rescaling forces the final result back into a decimal type throwing away the overflow logic. Something I've wanted to change for a while is to allow a (temporary) 32bit overflow which will also ultimately give some performance improvements.

I'll recreate this PR as an issue with the above updates for tracking.

PR #103, opened 2017-12-30
Clean up of div logic

paupino:
Interestingly enough this does slow it down slightly however I think it'll be easier to maintain going forward as well as find optimizations.


paupino:
Ideally I want to get division time down further however this at least provides a more maintainable solution (i.e. without the ambiguity). From here it should be easier to find optimizations (e.g. u32 divisor).

Issue #113, opened 2018-03-24
Panic when formatting precision on fractional number.


terrybrashaw:
**Version**: 0.7.2 (c22883b6cd16c78efa08fa8e0c34774d34a83ca9)

`Decimal` panics when formatting precision on a fractional number between `1 > n > -1` which contains leading zeros.
```rust
let foo: Decimal = "0.0023554701772169".parse().unwrap();
println!("{:.4}", foo);
```
> thread 'main' panicked at 'attempt to subtract with overflow', /home/terry/.cargo/git/checkouts/rust-decimal-5e2779474ac0f95f/c22883b/src/decimal.rs:1658:30

The expected output is:
> 0.0023

PR #114, opened 2018-03-24
Fix panic when formatting precision

terrybrashaw:
Fixes https://github.com/paupino/rust-decimal/issues/113


paupino:
Awesome, thank you for creating the corresponding issue too

PR #132, opened 2018-06-19
Optimization add/sub

xilec:
This changes simplify `add` function and make big speed up via agressive inlining functions of used for addition `Decimal`.
Actually, after adding `bench_dec_10k`, `bench_sum_10k` became show almost 1.75x worst result. Adding attribute `#[inline(always)]` for `add` and `sub` fixed it. I hope it have such positive effect in common, but I'm not sure about it.

Benchmarks:

before
```
test add_negative_pi         ... bench:         171 ns/iter (+/- 6)
test add_negative_point_five ... bench:          28 ns/iter (+/- 4)
test add_one                 ... bench:          29 ns/iter (+/- 4)
test add_one_hundred         ... bench:          29 ns/iter (+/- 1)
test add_pi                  ... bench:         166 ns/iter (+/- 11)
test add_point_zero_one      ... bench:          14 ns/iter (+/- 3)
test add_two                 ... bench:          29 ns/iter (+/- 4)
```
after
```
test add_negative_pi         ... bench:         164 ns/iter (+/- 4)
test add_negative_point_five ... bench:          21 ns/iter (+/- 3)
test add_one                 ... bench:          26 ns/iter (+/- 1)
test add_one_hundred         ... bench:          26 ns/iter (+/- 3)
test add_pi                  ... bench:         164 ns/iter (+/- 22)
test add_point_zero_one      ... bench:          11 ns/iter (+/- 0)
test add_two                 ... bench:          26 ns/iter (+/- 4)
```

before
```
test sub_negative_pi         ... bench:         168 ns/iter (+/- 16)
test sub_negative_point_five ... bench:          24 ns/iter (+/- 4)
test sub_one                 ... bench:          35 ns/iter (+/- 2)
test sub_one_hundred         ... bench:          36 ns/iter (+/- 4)
test sub_pi                  ... bench:         173 ns/iter (+/- 12)
test sub_point_zero_one      ... bench:          22 ns/iter (+/- 3)
test sub_two                 ... bench:          35 ns/iter (+/- 2)
```
after
```
test sub_negative_pi         ... bench:         164 ns/iter (+/- 13)
test sub_negative_point_five ... bench:          21 ns/iter (+/- 1)
test sub_one                 ... bench:          27 ns/iter (+/- 0)
test sub_one_hundred         ... bench:          28 ns/iter (+/- 0)
test sub_pi                  ... bench:         164 ns/iter (+/- 13)
test sub_point_zero_one      ... bench:          12 ns/iter (+/- 0)
test sub_two                 ... bench:          27 ns/iter (+/- 2)
```

before
```
test bench_sum_10k           ... bench:      79,669 ns/iter (+/- 5,988)
```
after
```
test bench_sum_10k           ... bench:      41,334 ns/iter (+/- 1,217)
```

before
```
test bench_dec_10k           ... bench:     109,814 ns/iter (+/- 5,145)
```
after
```
test bench_dec_10k           ... bench:      46,768 ns/iter (+/- 2,222)
```

C# version .net core
```
     Method |     Mean |     Error |    StdDev |
----------- |---------:|----------:|----------:|
 DecDecimal | 109.2 us | 0.6073 us | 0.5384 us |
 SumDecimal | 112.0 us | 0.0913 us | 0.0854 us |
```


benches/lib_benches.rs
```diff
@@ -32,17 +32,20 @@ bench_decimal_op!(add_pi, +, "3.1415926535897932384626433832");
 bench_decimal_op!(add_negative_pi, +, "-3.1415926535897932384626433832");
 
 #[bench]
-fn bench_sum_10m(b: &mut ::test::Bencher) {
-    fn sum_10m(values: &[Decimal]) -> Decimal {
+fn bench_sum_10k(b: &mut ::test::Bencher) {
```
paupino: Looking at the two of these functions I'm thinking that `bench_add_10k` and `bench_sub_10k` may be better.

src/decimal.rs
```diff
@@ -1097,27 +1103,33 @@ fn sub_internal(value: &mut [u32], by: &[u32]) -> u32 {
     overflow
 }
 
+//#[inline]
 fn sub_part(left: u32, right: u32, overflow: u32) -> (u32, u32) {
```
paupino: Perhaps you could inline always this and reuse above?

src/decimal.rs
```diff
@@ -1097,27 +1103,33 @@ fn sub_internal(value: &mut [u32], by: &[u32]) -> u32 {
...
-        invert = true;
-    }
-
-    if part > overflow {
-        part -= overflow;
-    } else {
-        part -= overflow;
-        invert = true;
-    }
+    let part = 0x1_0000_0000u64 + u64::from(left) - (u64::from(right) + u64::from(overflow));
```
paupino: I'd probably feel a little better here with brackets around the first section to make the order of operations very blatantly obvious (i.e. `(0x1_0000_0000u64 + u64::from(left))`). 

In addition, to clean up it'd be a good idea to remove the superfluous comments below. We'll still have a copy of how it worked within Git if we need to revert.

I guess the last suggestion to round this off a bit further; it'd be great to ensure we have a test (or tests) that covers an extreme scenario at a part level. e.g. maybe 0XFFFF_FFFF_FFFF_FFFF - -0XFFFF_FFFF_FFFF_FFFF_FFFA or something like that. I've been trying a couple of examples on paper to test the logic but another addition to the test array would be ideal. 😄

src/decimal.rs
xilec: Actually, if we use `sub_part` in `sub3_internal`, it reduce performance about 12%. I suppose, we get speed down because of additional transforms `u32` values in `u64` and vice versa.

benches/lib_benches.rs
xilec: Done, as I understand

src/decimal.rs
xilec: Done

PR #141, opened 2018-07-01
Postgres write speed up by avoiding strings allocation

### LINKED ISSUE #82: Improve Postgres Write performance by avoiding string conversion
Currently writing to Postgres requires a string conversion. There may be an opportunity to avoid this conversion and consequently improve write performance.

xilec:
This changes fix #82 

before
```
test postgres::to_from_sql   ... bench:   2,170,957 ns/iter (+/- 29,919)
```

after
```
test postgres::to_from_sql   ... bench:     577,522 ns/iter (+/- 23,498)
```


src/postgres.rs
```diff
@@ -162,74 +164,52 @@ impl ToSql for Decimal {
...
-        let weight = if whole_portion.is_empty() {
-            -(decimal_portion.len() as i16)
+        let groups_diff = scale & 0x3;                           // groups_diff = scale % 4
+        let mut fractional_groups_count = (scale >> 2) as isize; // fractional_groups_count = scale / 4
+        fractional_groups_count += if groups_diff > 0 { 1 } else { 0 };
+
+        let mut mantissa = self.mantissa_array4();
+
+        if groups_diff > 0 {
+            let fractional_group_reminder = 4 - groups_diff;
```
paupino: Nit: should be `remainder`

src/postgres.rs
```diff
@@ -287,6 +267,23 @@ mod test {
...
+        "0.001",
+        "0.0001",
+        "0.00001",
+        "0.000001",
+        "1",
+        "-100",
+        "-123.456",
+        "119996.25",
+        "1000000",
+        "9999999.99999",
```
paupino: It'd be great to add those extra numbers on here too:

```
        "12340.56789",
        "79228162514264337593543950335", // 0xFFFF_FFFF_FFFF_FFFF_FFF_FFFF
        "4951760157141521099596496895", // 0x0FFF_FFFF_FFFF_FFFF_FFF_FFFF
        "4951760157141521099596496896", // 0x1000_0000_0000_0000_0000_0000
        "18446744073709551615",
        "-18446744073709551615"
```

`NUMERIC(26,6)` will be too small for some of those - one workaround is to just make it `NUMERIC` or to break the tests into two and have the second set use a larger PG scale.

src/postgres.rs
```diff
@@ -162,74 +164,52 @@ impl ToSql for Decimal {
...
-                let mut x = x.to_owned();
-                while x.len() < 4 {
-                    x.push('0');
-                }
-                x.into_iter().collect::<String>()
-            })
-            .collect::<Vec<String>>();
-        let weight = if whole_portion.is_empty() {
-            -(decimal_portion.len() as i16)
+        let groups_diff = scale & 0x3;                           // groups_diff = scale % 4
```
paupino: Clever!

src/postgres.rs
xilec: done

Issue #71, opened 2017-11-21
Unexpected Panics


rkusa:
Hi,

I understand that rust-decimal panics in case of unrecoverable errors. However, Today I encountered two panics that seem odd.

```rust
    let rate = Decimal::new(19, 2);
    let part = rate / (rate + Decimal::new(1, 0));
    let result = Decimal::new(169, 0) * part; // panics
```

This will panic with `Multiplication underflowed`. However if I switch the operands (to `part * Decimal::new(169, 0)`), it does not panic anymore.

I tried to also fix this issue by rounding the `part` variable first, but this panicked as well:

```rust
    let rate = Decimal::new(19, 2);
    let part = rate / (rate + Decimal::new(1, 0));
    let part = part.round_dp(2); // panics
```

It panics with `Cannot have an exponent greater than 28 (29 > 28)`.

Thanks

paupino:
I think this is due to an incorrect assumption I made in terms of underflow vs rounding (i.e. here https://github.com/paupino/rust-decimal/blob/master/src/decimal.rs#L1624). Let me reproduce this in a test and see what I can do to fix it.

paupino:
(i.e. We're not losing precision if they're 0's)

paupino:
I've discovered the bug: it's in the division routine whereby we get an underflow in the division however don't attempt to resolve it there. I understand the problem now so will look to get out a solution as soon as I can. The result SHOULD be scale 9 however because we set underflow to true we keep it in an intermediary form (29). This is a shared issue in both simply because division _silently_ allows invalid scales at the moment.

rkusa:
Wow, Thanks a lot @paupino for investigating and fixing the issue!

PR #142, opened 2018-07-03
Add additional rounding strategies

### LINKED ISSUE #140: Add support for additional rounding strategies
Hi, first of all thanks for the library,I just started using it and the API is excellent, really easy to get started with :slightly_smiling_face: 

For my use case, I need to make use of the round half up rounding strategy, I was wondering if you were open to adding the ability to either:
 - use more than banker's rounding
 - provide a function that acts as the rounding strategy

I'm more than willing to attempt implementing this, I just wanted to see if it was something you were open to adding into the crate?

AndrewSpeed:
This PR implements the rounding strategies discussed in #140, namely `RoundingStrategy::RoundHalfUp` and `RoundingStrategy::RoundHalfDown`.

In order to do so, it maps the existing strategy to `RoundingStrategy::BankersRounding`.

Resolves #140

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -555,13 +555,13 @@ impl Decimal {
             };
         }
 
+        let old_scale = self.scale();
+
         // return early if decimal has a smaller number of fractional places than dp
         // e.g. 2.51 rounded to 3 decimal places is 2.51
-        if self.scale() <= dp {
-            return *self
+        if old_scale <= dp {
+            return *self;
         }
-
-        let old_scale = self.scale();
 
         let mut value = [self.lo, self.mid, self.hi];
         let mut value_scale = self.scale();
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -105,10 +105,13 @@ pub struct Decimal {
 ///
 /// `RoundingStrategy::BankersRounding` - Rounds toward the nearest even number, e.g. 6.5 -> 6, 7.5 -> 8
 /// `RoundingStrategy::RoundHalfUp` - Rounds up if the value >= 5, otherwise rounds down, e.g. 6.5 -> 7,
+/// `RoundingStrategy::RoundHalfDown` - Rounds down if the value =< 5, otherwise rounds up, e.g.
+/// 6.5 -> 6, 6.51 -> 7
 /// 1.4999999 -> 1
 pub enum RoundingStrategy {
     BankersRounding,
-    RoundHalfUp
+    RoundHalfUp,
+    RoundHalfDown
 }
 
 #[allow(dead_code)]
@@ -633,6 +636,14 @@ impl Decimal {
                     }
                     Ordering::Greater => {
                         // Doesn't matter about the decimal portion
+                        add_internal(&mut value, &ONE_INTERNAL_REPR);
+                    }
+                    _ => {}
+                }
+            },
+            RoundingStrategy::RoundHalfDown => {
+                match order {
+                    Ordering::Greater => {
                         add_internal(&mut value, &ONE_INTERNAL_REPR);
                     }
                     _ => {}
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -100,6 +100,17 @@ pub struct Decimal {
     mid: u32,
 }
 
+/// `RoundingStrategy` represents the different strategies that can be used by
+/// `round_dp_with_strategy`.
+///
+/// `RoundingStrategy::BankersRounding` - Rounds toward the nearest even number, e.g. 6.5 -> 6, 7.5 -> 8
+/// `RoundingStrategy::RoundHalfUp` - Rounds up if the value >= 5, otherwise rounds down, e.g. 6.5 -> 7,
+/// 1.4999999 -> 1
+pub enum RoundingStrategy {
+    BankersRounding,
+    RoundHalfUp
+}
+
 #[allow(dead_code)]
 impl Decimal {
     /// Returns a `Decimal` with a 64 bit `m` representation and corresponding `e` scale.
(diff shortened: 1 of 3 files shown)


src/decimal.rs
```diff
@@ -514,116 +528,168 @@ impl Decimal {
...
+
+        // If the decimal_portion is zero then we round based on the other data
+        let mut cap = [5, 0, 0];
+        for _ in 0..(old_scale - dp - 1) {
+            mul_by_u32(&mut cap, 10);
+        }
+        let order = cmp_internal(&decimal_portion, &cap);
+
+
+        match strategy {
```
AndrewSpeed: I'm thinking about trying to refactor strategies into a trait, that way this crate can provide some default implementations, but users can extend it further if they wish to do so.

If this approach is taken, I think some additional methods would need to be exposed to make it possible for alternative strategies to be implemented.

tests/decimal_tests.rs
```diff
@@ -648,6 +648,275 @@ fn it_can_round_to_2dp() {
     assert_eq!("6.12", b.to_string());
 }
 
+#[test]
+fn it_can_round_to_2dp_using_bankers_rounding_strategy() {
+    let a = Decimal::from_str("6.12345").unwrap();
```
paupino: On a side note, I've been more inclined to use test arrays going forward simply because it makes it easier to add tests. I think you could do probably set up something similar here to set precedent:
```
using RoundingStrategy::*;
let tests = &[
    ("6.12345", 2, BankersRounding, "6.12"),
    ("6.126", 2, BankersRounding, "6.13"),
    // etc
];
for &(input, dp, strategy, expected) in tests {
    let a = Decimal::from_str(input).unwrap();
    let b = a.round_dp_with_strategy(dp, strategy);
    assert_eq!(expected, b.to_string());
}
```
The odd one out of course is the test marked `// Issue #71`. For this one it could have a separate combined test which iterates the various rounding strategies on this irrational number.

Of course I realize that you're reusing an existing pattern - it's just one I'd like to move away from at some stage in the future. Whether that's now or later is up for debate 😄

src/decimal.rs
```diff
@@ -514,116 +528,168 @@ impl Decimal {
...
-            // Short circuit for zero
-            if self.is_zero() {
-                return Decimal {
-                    lo: 0,
-                    mid: 0,
-                    hi: 0,
-                    flags: flags(self.is_sign_negative(), dp),
-                };
-            }
+        let old_scale = self.scale();
```
paupino: Any reason you do this after the previous comparison? As it is it just means the flags get evaluated twice.

src/decimal.rs
paupino: Interesting idea. I like the extensibility aspect of this - how would you ensure it remained ergonomic?

AndrewSpeed: Assuming you're talking about ergonomics for the method and/or trait, my thought was that we could have this method call out to a method defined on the trait e.g. `round_to_dp(n: u32)`, from there it's up to the strategy on _how_ to do the rounding.

This should give the solution the most flexibility, as the implementation won't know anything about the rounding, except that it returns a `Decimal`.

In summary, something like:
```rust
trait RoundingStrategy {...}

pub fn round_dp_with_strategy(&self, dp: u32, strategy: RoundingStrategy) -> Decimal {
  strategy.round_to_dp(self, dp)
}
```

Can you see any problems and/or concerns with that approach?

paupino: Playing it out from an integration perspective; are you thinking:
```
let number = dec!(1.2345);
let _ = number.round_with_strategy(2, BankersRoundingStrategy::default());
```

You could also feasibly invert this relationship also:
```
let number = dec!(1.2345);
let bankers_rounding = BankersRounding::with_dp(2);
let _ = bankers_rounding.round(number);
```

The problem I see with each is accessing/setting the internals effectively for efficiency. I think that at the moment, an `enum` is great for optimizing towards efficiency because it allows everything to be handled within the same scope. If another rounding strategy is required later on then it is feasible for someone to do this using the public API of the decimal type in an approach similar to the second example (at a small perf cost).

On a side note, the function name is starting to get quite long and since it'll become part of the public API is the perfect time to think about it. I've been trying to think of alternatives to assist with this but the best I can come up with is _maybe_ dropping the `_dp`. I'm still a little torn here... I'm definitely open to keeping it the same but if you think of something more succinct then I'm all ears!

AndrewSpeed: That's along the lines of what I was thinking of, however another alternative would be to set the rounding strategy when the decimal is created. `BankersRounding` would be the default, then the user could update it or use a specialised constructor like:

```rust
let number = Decimal::new_with_strategy(134, 2, Rounding::RoundHalfUp);
let rounded = number.round_dp(2);
```

This would eliminate the naming problem entirely 🙂 

I can definitely see where you're coming from, it'd be a trade-off of performance for flexibility, I'm just not sure what that impact would be, and whether it'd be worth it in the long run.

AndrewSpeed:
Sorry for the delay in updating this PR @paupino I've added the feedback you requested. Let me know if there's anything else that you'd like changed :slightly_smiling_face:

paupino:
Thank you @AndrewSpeed ! I see you refactored the `BankersRounding` and `RoundHalfUp` tests to use the array style which is looking great. `RoundHalfDown` tests are still in the old format though - could these be also switched over?

AndrewSpeed:
:+1: Will have a commit up to address that shortly

paupino:
Awesome, thank you for bearing with me through that PR! I'll merge these once the tests pass.

AndrewSpeed:
No problem at all, sorry for the delay between feedback and it being addressed, life happened and I got bogged down in other things :sweat_smile:

PR #147, opened 2018-09-14
Support const_fn

neoeinstein:
Makes a few of the decimal functions `const fn`, allowing the compiler to perform the appropriate action at compile time instead of needing to delay until runtime. Requires the nightly compiler and the `const_fn` feature gate. The functionality is gated behind the `const_fn` feature on this crate.


paupino:
Thanks for the contribution! 

It'd be great to capture the performance improvements within the PR that you've found from this change for reference. I've been following this feature quite closely also :) 

Since this changes behavior it'd also be useful to change the build so that we build with the feature and without the feature on nightly. Also, at the moment this feature can't co-exist with the `postgres` feature since this uses `lazy-static`. 

I'm happy to help with the build stuff if you want a hand getting this PR ready for merge.

paupino:
I'll relook at this PR early Dec once Rust 2018 lands. This feature will be stable so will try to get it in asap.

neoeinstein:
Now that 1.32 has landed, want me to clean this up to land? Probably fine to keep it behind a feature switch, or I can add a build step that automatically enables the feature of the compiler supports it.

paupino:
@neoeinstein That would be great thank you! Either way is ok with me.

neoeinstein:
Ok. This has been updated, and I've rebased on the current master. Tests pass, and I fixed the postgres feature as well. It now can build its table at compile time rather than as a lazy static. A test was added to verify that the resulting constant array matches the existing lazy-static array (I use `from_parts` for all of them as `new` can't yet be "`const`-ified".

PR #152, opened 2018-11-30
Fix parsing when underscore separators precede a decimal point

barnardb:
This fixes the parsing of numbers like "1_234.56" which have an "_" separator followed later by a decimal point. Currently this number is incorrectly parsed as equivalent to "12345.6" instead of "1234.56".

The incorrect value arises because the determination of the scale in `Decimal::from_str` made reference to the difference between the offsets of the start of the number and the decimal point, causing "_" separators to incorrectly affect the scale.

This PR fixes the problem by changing `Decimal::from_str` to remember the number of digits preceding the decimal point, rather than offsets.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1501,8 +1501,7 @@ impl FromStr for Decimal {
         }
 
         // should now be at numeric part of the significand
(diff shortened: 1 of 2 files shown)


paupino:
Great catch - thank's for your help!

PR #154, opened 2019-01-04
Use proc-macro-hack to allow dec! in expr position

neoeinstein:
Added a new crate `rust_decimal_macro_impls` to hold the implementation,
re-exporting the procedural macro through `rust_decimal_macros` in order
to fit the pattern for using the `proc-macro-hack` crate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ before_script:
   - "./.travis/setup.sh"
 
 script:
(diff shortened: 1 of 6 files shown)


neoeinstein:
If merged, this will fix #11.

paupino:
Thank you for this. I want to look into the hack to understand what it does before merging - I'll get around to this ASAP.

PR #160, opened 2019-03-11
serde: try from_scientific when from_str fails

mattjbray:
Allows deserializing strings in scientific notation.


paupino:
Thanks @mattjbray for the contribution!

PR #163, opened 2019-03-25
Add num trait prelude

### LINKED ISSUE #162: Re-exports the implemented num traits
Currently there is no way to use the `FromPrimitive`, `One`, `ToPrimitive` and `Zero` traits implemented by `Decimal` without separately depending on the `num` crate (at least from what I understand). This could potentially cause hard to debug incompatibility issues if someone depends on a different version of the `num` crate than the one used by `rust-decimal`.

Adding a prelude that re-export these traits would do the trick.

jean-airoldie:
This fixes #162.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,3 +64,7 @@ mod serde_types;
 
 pub use decimal::{Decimal, RoundingStrategy};
 pub use error::Error;
+
+pub mod prelude {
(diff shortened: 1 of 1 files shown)


paupino:
This is great; I could feasibly foresee exposing further types in the future also (e.g. `FromStr` is a common trait that is reused, though it doesn't suffer the same versioning constraints).

PR #165, opened 2019-03-25
Edition 2018

### LINKED ISSUE #164: Update to edition 2018
This would make the crate more readable because of the module system improvements.

jean-airoldie:
Fixes #164. Also ran `cargo fmt`.


src/decimal.rs
```diff
@@ -52,43 +57,42 @@ const MIN: Decimal = Decimal {
...
-        1_000_000,
-       10_000_000,
-      100_000_000,
+    1,
+    10,
+    100,
+    1_000,
+    10_000,
+    100_000,
+    1_000_000,
```
paupino: This is one part I disagree with cargo fmt. I'm a big right aligned numbers guy!

src/postgres.rs
```diff
@@ -1,16 +1,16 @@
...
-use std::fmt;
-use std::io::Cursor;
-use std::result::*;
-use decimal::{div_by_u32, is_all_zero};
-use decimal::mul_by_u32;
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+
+use num::Zero;
+
+use crate::Decimal;
```
paupino: TODO: I suspect it's breaking these apart (similar in another file above) due to ordering issues. We should probably group crate vs std vs external.

macro_impl/src/lib.rs
```diff
@@ -1,12 +1,13 @@
 extern crate proc_macro;
-extern crate proc_macro_hack;
-extern crate rust_decimal;
-extern crate quote;
 
 use proc_macro::TokenStream;
+
```
paupino: This seems a little overkill - perhaps because of grouping.

PR #177, opened 2019-04-24
Implement checked operations

### LINKED ISSUE #80: Implement `checked` operator functions
To help alleviate some of the panic scenario's currently present in the library we should introduce `checked` methods. Checked returns an `Option<Decimal>` and returns `None` on overflow.

0e4ef622:
Closes #80.

I just copy/pasted the code, replaced the panics, and reimplemented the unchecked operations using the checked variant. Subtraction will now panic with `Subtraction overflowed` instead of `Addition overflowed`.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1056,12 +1056,435 @@ impl Decimal {
             flags: flags(!positive, -exponent10 as u32),
         })
     }
+
+    /// Checked addition. Computes `self + other`, returning `None` if overflow occurred.
+    #[inline(always)]
+    pub fn checked_add(self, other: Decimal) -> Option<Decimal> {
+        // Convert to the same scale
+        let mut my = [self.lo, self.mid, self.hi];
+        let mut my_scale = self.scale();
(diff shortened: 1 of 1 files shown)


src/decimal.rs
```diff
@@ -2217,239 +2572,72 @@ impl<'a> AddAssign<&'a Decimal> for Decimal {
...
+forward_all_binop!(impl Mul for Decimal, mul);
+
+impl<'a, 'b> Mul<&'b Decimal> for &'a Decimal {
+    type Output = Decimal;
+
+    #[inline]
+    fn mul(self, other: &Decimal) -> Decimal {
+        match self.checked_mul(*other) {
+            Some(prod) => prod,
+            None => panic!("Multiplication overflowed"),
```
paupino: I like the idea of being more explicit with the errors (similar to `div` below) but that being said, that is more of an implementation issue that we can evolve towards. Accordingly, as we evolve we could probably keep the message attached to the result too (e.g. `imp Display`).

Issue #186, opened 2019-04-30
Should `const_fn` be a default feature now?


0e4ef622:
It works on stable.

devashishdxt:
Created a PR for this: https://github.com/paupino/rust-decimal/pull/189

PR #189, opened 2019-05-08
Makes `const_fn` a default feature

devashishdxt:
Details:

- Removes `const_fn` and `lazy_static` features from feature list.
- Defaults all branches of conditional compilation with `const_fn` route.
- Removes `lazy_static` dependency.

Fixes https://github.com/paupino/rust-decimal/issues/186.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ travis-ci = { repository = "paupino/rust-decimal", branch = "master" }
 [dependencies]
 num = "0.2"
 byteorder = "1.3"
-lazy_static = { version = "1.3", optional = true }
 postgres = { version = "0.15", optional = true }
 serde = { version = "1.0", optional = true }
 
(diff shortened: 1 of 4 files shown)


devashishdxt:
@paupino: Can you please review this?

paupino:
The changes look good. I wonder however if the feature flag should instead be reversed - that is, enabled by default but still available (similar to `no_std`) so that older versions of the Rust compilers can still use this? That way, by default we can keep adding new compiler features to the library but at the same time still support older versions of Rust.

Let me think about that for a little, but also curious to hear your thoughts.

devashishdxt:
I think instead of using features, we can use [select-rustc](https://crates.io/crates/select-rustc) to define default behaviour for different Rust versions.

paupino:
Will likely just provide a minimum Rust Compiler version for now (i.e. https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1310-2018-12-06). For proper backwards compatible support I'd also need to address the build pipeline which may be something I can introduce in the future.

PR #191, opened 2019-05-08
Diesel support

vorot93:
Postgres conversion code is refactored to be library-independent with Diesel and rust-postgres support bolted on top of it. Tests are taken from Diesel's bigdecimal support, but I suspect that more exhaustive testing is required.

Should only be merged after #187.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ travis-ci = { repository = "paupino/rust-decimal", branch = "master" }
 [dependencies]
 num = "0.2"
 byteorder = "1.3"
+diesel = { version = "1.4", default-features = false, features = ["postgres"], optional = true }
 postgres = { version = "0.15", optional = true }
 serde = { version = "1.0", optional = true }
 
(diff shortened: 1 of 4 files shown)


paupino:
@vorot93 This PR is still in draft. Just wanted to check if that was intended?

vorot93:
It is. I never had the time to actually test this in production setting. It should be working but I'm not 100% sure. If somebody else can test it or it is not important I can mark it as ready for review.

neysofu:
Hi, any updates on the situation? Please let me know if I can help in any way.

vorot93:
Marking it ready for review. I don't use it with diesel right now but if anyone starts testing this out, I will happily merge fixes and do regular rebases on top of current master.

neysofu:
I'll be giving it a look and run some more tests in the following days.

neysofu:
We haven't met any other bugs so far.

paupino:
Merging this to master - we'll get this into the next minor edition update.

paupino:
Thank you both @neysofu and @vorot93 for your help!

PR #195, opened 2019-06-17
quick and dirty change of using `postgres` library to `tokio-postgres`.

kaibyao:
I've done the minimal work of making `rust-decimal` compatible with the latest release candidate of tokio-postgres (v0.4.0-rc2 as of now).

There's still compilation warnings from clippy, but everything seems to work.


kaibyao:
Need to fix my tests, will resubmit later.

PR #198, opened 2019-07-10
Upgrade to rust-postgres 0.16

pimeys:
An upgrade path to postgres 0.16.0 is a bit bumpy and we're using it (and soon the async version) already, so we needed to upgrade rust-decimal before postgres was stabilized... Don't know really when do you want to merge this and it is a breaking change for the users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ travis-ci = { repository = "paupino/rust-decimal", branch = "master" }
 [dependencies]
 num = "0.2"
 byteorder = "1.3"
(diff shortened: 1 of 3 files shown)


paupino:
Thanks for the contribution! I'm currently waiting for `postgres` to come out of RC - but that being said, we could consider making this a feature if it's useful for others? i.e. postgres_next or something.

pimeys:
Would be nice. We at Prisma would like to publish our query AST / connector crate and this dependency kind of prevents us for doing so. Maybe next week I'll update the PR for that.

-------- Original Message --------

paupino:
Merged in #206

PR #196, opened 2019-06-17
postgres => tokio-postgres

kaibyao:
I've done the work of making rust-decimal compatible with the latest release candidate of tokio-postgres (v0.4.0-rc2 as of now).

PR #203, opened 2019-10-30
Support tokio-postgres 0.5.0 and async/await

pimeys:
Async/await is soon stable and we should support the async/await version of tokio-postgres.

This PR was a bit more involved due to tokio-postgres now using `bytes` in the `to_sql` function.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,16 +22,16 @@ num = "0.2"
 byteorder = "1.3"
 float-cmp = "0.5.0"
 diesel = { version = "1.4", default-features = false, features = ["postgres"], optional = true }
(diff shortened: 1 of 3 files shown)


paupino:
Thank you for this. Similarly related is #196.
My main hesitation for merging is because I'm waiting for the stable release. It may be possible however that I could create a similar feature flag for including alpha/beta features.

pimeys:
It would help us to release [quaint](https://github.com/prisma/quaint) to crates.io finally :)

ohmyarch:
@pimeys Please adapt to tokio 0.2.1 which uses bytes 0.5.0.

pimeys:
PR updated to tokio 0.2.1 and bytes 0.5.

PR #210, opened 2020-01-22
Retain trailing zeroes received from Postgres

hathawsh:
For correctness, this crate should retain trailing zeroes in numeric values received from Postgres. This patch fixes that.


paupino:
Thanks @hathawsh for the contribution! In order to get this merged there are some tests that also need updating. Fortunately, the tests use a shared array defining actual/expected values. e.g. this test decimal needs the expected value to be updated to have 6DP: https://github.com/paupino/rust-decimal/blob/master/src/postgres.rs#L551
Are you able to resolve? Thanks!

hathawsh:
I resolved the test failures (I had forgotten to use "--all-features" in cargo test) and added support for an environment variable, POSTGRES_URL, that lets me connect to Postgres in a way that makes sense in my environment.

PR #211, opened 2020-01-22
Correct hash calculation by normalizing before hashing

hathawsh:
The Rust Hash documentation states that Hash implementations must maintain this invariant:

    k1 == k2 -> hash(k1) == hash(k2)

rust_decimal currently fails that invariant when trailing zeroes are involved; this patch fixes that. Test included.


tests/decimal_tests.rs
```diff
@@ -1332,3 +1332,60 @@ fn to_from_sql() {
...
+    assert_eq!(k1, k3);
+    assert_ne!(k1, k4);
+
+    let h1 = hash_it(k1);
+    let h2 = hash_it(k2);
+    let h3 = hash_it(k3);
+
+    assert_eq!(h1, h2);
+    assert_eq!(h1, h3);
+
```
paupino:
```suggestion
assert_ne!(h1, h4);
```

hathawsh:
Good suggestion. Done.

PR #218, opened 2020-01-30
Change set_sign so that it is more readable

paupino:
Current api to update a decimal so that it is negative:
```
let mut a = Decimal::new(123, 2);
a.set_sign(false);
```

New api to update a decimal so that it is negative:
```
let mut a = Decimal::new(123, 2);
a.set_sign(Sign::Negative);
```


paupino:
Closing this. Will implement `set_sign` in v2. For now, will make it `set_sign_positive`

PR #227, opened 2020-03-04
Some improvements on Decimal::from_scientific()

edwardycl:
I tried to parse a scientific notation with 'E' in it (e.g. "1.23E-4") and it failed, so I changed the function to allow uppercase 'E' and made a few other changes.
- Allow uppercase 'E' in scientific notation
- Allow scientific notation in `dec!` macro
- Remove string allocation in `from_scientific` function

Also, I added `Decimal::from_str()` benchmark because I'm working on it currently. It's already ~20% faster, but handling the newly added radix is taking some time. I'll make a new PR when ready.


paupino:
Awesome, thank you for your contribution!

Issue #234, opened 2020-04-09
Optimizing to_f64?


hengchu:
First of all, I want to thank you for this wonderful package!

I have an data-intensive application that utilizes `rust-decimal`, and calls `to_f64()` a lot. A recent instrumentation showed me that my application is spending roughly 10% of the time in `to_f64()` due to the current conversion algorithm (https://github.com/paupino/rust-decimal/blob/ba5ba1f661bf67a8e12bba8c9fdc438d69371e69/src/decimal.rs#L2526-L2541), which essentially converts `Decimal` to `f64` through an intermediate string.

The current implementation has a TODO that suggests there may be a faster way of doing this conversion. I'd love to help with this optimization. Do you have some pointers on how to get started? Would the mantissa algorithm be something like the inverse of `Decimal::base2_to_decimal`?

paupino:
Ah yes! That wouldn't be very performant at all.

Essentially, we'd want to be able to reconstruct the raw float again similar to how it's deconstructed here: 
https://github.com/paupino/rust-decimal/blob/master/src/decimal.rs#L2457-L2460

To do this, we'd need to recalculate the parts of the formula: `n = (sign*-1) * 2^exp * mantissa`
Sign is easy - the tricky parts are firstly calculating the `exponent2` and the `mantissa`. I  think you could do this in a very similar way (except, the inverse) to how it's been leveraged within this function.

I'm more than happy to pick it up also, but am more than grateful to get some help for maintenance - so thank you!

paupino:
Closed with #235

PR #235, opened 2020-04-09
implemented an optimized to_f64()

hengchu:
This is a follow up of the issue here: https://github.com/paupino/rust-decimal/issues/234

I ended up implementing an alternative (IMHO simpler) idea. Hopefully the code is readable enough.

I thought about re-constructing the f64 from scratch using the Decimal's internal parts, but quickly ran into some questions that I'm not sure how to address. The gist is this:

We know a Decimal value is `sign * m / 10^e`, where m is a 96-bit integer, and e is at most 28, while an f64 value is `sign * m / 2^e`. This means we need to convert a base 10 exponent into a base 2 exponent.

So, we can rewrite the Decimal value `sign * m / 10^e` into `sign * (m / 5^e) / 2^e`. Let `m' = m / 5^e`. This `m'` will be the mantissa for the f64 value. The problem is `m` itself is not necessarily divisible by 5, the best we can do is approximate it with the integer that's closest to `m / 5^e`. So there's some loss of precision.

So, IMO, since we are losing precision here anyways, maybe it's worth it to simply use `f64` arithmetic to calculate the fractional part of the Decimal value directly, and add that to the integral part of the Decimal value. Hence, the alternative algorithm below.

This implementation breaks one of the test cases in `it_converts_to_f64` with the following output:
```
---- it_converts_to_f64 stdout ----
thread 'it_converts_to_f64' panicked at 'assertion failed: `(left == right)`
  left: `0.0000000000025`,
 right: `0.0000000000025000000000000007`', tests/decimal_tests.rs:1065:5
```

 The test case is originally doing exact floating number comparison between the LHS and RHS. I revised it to use approximate comparison through the `float-cmp` crate. This shows the `LHS` and `RHS` are less than 2 ULPS apart, which means they are less than 2 units of precision apart at their current exponent.

This may or may not be acceptable for you... I'm happy to consider other routes that can potentially avoid modifying this existing test case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ serde_json = "1.0"
 serde_derive = "1.0"
 tokio = { version = "0.2", features = ["rt-threaded", "test-util", "macros"] }
 futures = "0.3"
+float-cmp = "0.6"
 
 [features]
 default = ["serde"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ serde_json = "1.0"
 serde_derive = "1.0"
 tokio = { version = "0.2", features = ["rt-threaded", "test-util", "macros"] }
 futures = "0.3"
-float-cmp = "0.6"
 
 [features]
 default = ["serde"]
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2531,11 +2531,17 @@ impl ToPrimitive for Decimal {
                 None => None,
             }
         } else {
-            // TODO: Utilize mantissa algorithm.
-            match self.to_string().parse::<f64>() {
-                Ok(s) => Some(s),
-                Err(_) => None,
-            }
+            let sign: f64 = if self.is_sign_negative() { -1.0 } else { 1.0 };
+            let mut mantissa: u128 = self.lo.into();
+            mantissa |= (self.mid as u128) << 32;
+            mantissa |= (self.hi as u128) << 64;
+            // scale is at most 28, so this fits comfortably into a u128.
+            let scale = self.scale();
+            let precision: u128 = 10_u128.pow(scale);
+            let integral_part = mantissa / precision;
+            let frac_part = mantissa % precision;
+            let frac_f64 = (frac_part as f64) / (precision as f64);
+            Some(sign * ((integral_part as f64) + frac_f64))
         }
     }
 }
--- a/tests/decimal_tests.rs
+++ b/tests/decimal_tests.rs
@@ -7,6 +7,8 @@ use std::{
     str::FromStr,
 };
 
+use float_cmp::approx_eq;
+
 // Parsing
 
 #[test]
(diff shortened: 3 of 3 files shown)


paupino:
This makes sense and seems like a reasonable trade-off. Let me run a few scenarios locally when I get a chance before merging.

paupino:
I was wondering about overflow/underflow handling in this which got me writing a few tests. Oddly, I notice that I could not replicate the approximate comparison issue that you listed above. What platform are you using? 

I'm curious also whether these tests pass/fail on your machine? 

```rust
    assert_eq!(
        0.25e-25_f64,
        Decimal::from_str("0.000000000000000000000000025").unwrap().to_f64().unwrap(),
    );
    assert_eq!(
        2.1234567890123456789012345678_f64,
        Decimal::from_str("2.1234567890123456789012345678").unwrap().to_f64().unwrap(),
    );
```

The original test, just out of interest - compiles and tests just fine as:
```rust
    assert_eq!(
        0.25e-11_f64,
        Decimal::from_str("0.0000000000025").unwrap().to_f64().unwrap(),
    );
```

All in all, I think the `float-cmp` library is correct and what we should be using - I'm more curious as to the differing results to make sure it's captured in the CI process!

paupino:
Side note; I'm quite convinced the approach you've taken is going to be more correct than the initial discussion in #234. I'm keen to get this merged - let's add the underflow tests above to make sure they work ok as well as an overflow test (below) and we should be good to go.

```rust
assert_eq!(
        None,
        // Cannot be represented in an f64
        Decimal::from_str("21234567890123456789012345678").unwrap().to_f64(),
    );
```

hengchu:
> I was wondering about overflow/underflow handling in this which got me writing a few tests. Oddly, I notice that I could not replicate the approximate comparison issue that you listed above. What platform are you using?
> 
> I'm curious also whether these tests pass/fail on your machine?
> 
> ```rust
>     assert_eq!(
>         0.25e-25_f64,
>         Decimal::from_str("0.000000000000000000000000025").unwrap().to_f64().unwrap(),
>     );
>     assert_eq!(
>         2.1234567890123456789012345678_f64,
>         Decimal::from_str("2.1234567890123456789012345678").unwrap().to_f64().unwrap(),
>     );
> ```

These two do pass just fine.

> The original test, just out of interest - compiles and tests just fine as:
> 
> ```rust
>     assert_eq!(
>         0.25e-11_f64,
>         Decimal::from_str("0.0000000000025").unwrap().to_f64().unwrap(),
>     );
> ```
> 
> All in all, I think the `float-cmp` library is correct and what we should be using - I'm more curious as to the differing results to make sure it's captured in the CI process!

This is very strange... Now the original test passes for me as well. I'm on i7 Haswell. The only difference between now and then was I was running a CPU-bound job on one of the cores... But now even if I do spin up a core I can no longer reproduce the failure. Very odd.

I will add the overflow/underflow tests now.

hengchu:
I reverted changes to the original test case for now... Let's see if CI still passes?

paupino:
Looks like it passed! Thank you for your help here. I'm still super curious what caused the precision error earlier...
Anyway, let's merge this and evolve from there.

hengchu:
I'm very curious as well. If anything pops up again, please ping me :) Thanks for the quick responses.

paupino:
As a side note; some benches need to be added around all of these conversions. But that can be another PR!

PR #238, opened 2020-05-07
Added RoundingStrategy::{RoundUp, RoundDown}

jean-airoldie:
The strategy `RoundUp` and `RoundDown` allow the user to unconditionally
round up or down, respectively.

I think this is technically a breaking change.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -119,10 +119,14 @@ pub struct Decimal {
 /// `RoundingStrategy::RoundHalfDown` - Rounds down if the value =< 5, otherwise rounds up, e.g.
 /// 6.5 -> 6, 6.51 -> 7
 /// 1.4999999 -> 1
+/// `RoundingStrategy::RoundDown` - Always round down.
+/// `RoundingStrategy::RoundUp` - Always round up.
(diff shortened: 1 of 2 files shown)


src/decimal.rs
```diff
@@ -771,6 +775,12 @@ impl Decimal {
                     _ => {}
                 }
             }
+            RoundingStrategy::RoundUp => {
+                if decimal_portion != [0, 0, 0] {
```
paupino, line 779: For consistency, we should probably use `is_all_zero`
```suggestion
                if is_all_zero(&decimal_portion) {
```

paupino: Ah sorry - I got the logic for this wrong. It should be `!is_all_zero`!

src/decimal.rs
```diff
@@ -771,6 +775,12 @@ impl Decimal {
                     _ => {}
                 }
             }
+            RoundingStrategy::RoundUp => {
+                if is_all_zero(&decimal_portion) {
```
paupino, line 779:
```suggestion
                if !is_all_zero(&decimal_portion) {
```

PR #244, opened 2020-05-14
Fix edge case in from_i128_with_scale

0e4ef622:
`Decimal::from_i128_with_scale(i128::MIN, 0)` would incorrectly return `-0`. This was due to the fact that `i128::MIN.wrapping_neg() == i128::MIN`.


paupino:
Thank you for submitting a fix for this issue. Can you also capture the fixed issue with an additional test case? Ideally the test would fail before applying this patch, but succeed after applying the patch.

0e4ef622:
Ok, I added the test case

PR #261, opened 2020-08-02
Consolidate duplicated zero checks in cmp

KiChjang:
Just a small refactor here to dedupe common code fragments.


paupino:
Looks good, thank you!

PR #263, opened 2020-08-17
Add support for no_std

### LINKED ISSUE #190: Add `no_std` support
Add support using this crate in `no_std` environments. I think this crate is the desired solution in many environments with limited memory and requirement of good performance.

c410-f3r:
Fixes #190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,13 +12,13 @@ categories = ["science","data-structures"]
 license = "MIT"
 
 [dependencies]
-num-traits = "0.2"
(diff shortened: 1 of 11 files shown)


paupino:
This seems to be good, though I want to run it through the wringer a little bit to see how it holds up. The one thing I'm wondering about (and just want to confirm) is whether the `defaults` section should include `std`. It seems to work fine without it in this PR so just want to see how it works in a few other projects I have running.

c410-f3r:
`std` isn't needed that all and the only hard requirement is the built-in `alloc` crate

PR #264, opened 2020-08-26
Fix bench build

c410-f3r:
Building of the `to_from_sql` benchmark is currently broken


paupino:
Thank you. I should really add this back into the CI - at least from a build perspective. I wasn't getting meaningful results running it during the CI pipeline, however ensuring it continues building is important.

PR #280, opened 2020-10-18
Implemented sum trait for immutable reference of decimals

xilec:
before this PR only this case possible
```rust
let sum : Decimal = vs.iter().cloned().sum();
```
after this PR we can use
```rust
let sum : Decimal = vs.iter().sum();
```
By my local benchmark, last variant about 2% faster than first variant, but it in bounds of precision.


paupino:
Looks good!

Issue #276, opened 2020-09-17
Division prohibitively slow


smessmer:
Thank you very much for this great project, the accuracy offered by this over f64 is very helpful to me.

I'm using rust-decimal 1.8.1.

Unfortunately, it turned out that I cannot use the division operator for decimals because it is prohibitively expensive. Is there a way to optimize it and get closer to the performance of multiplication?

**Benchmark:**
```
$ cargo run --release
    Finished release [optimized] target(s) in 0.01s
     Running `target/release/mytest`
rust-decimal division : 9388.19 ns
rust-decimal multiplication : 36.4715 ns
f64 division : 1.2322 ns
f64 multiplication : 1.4745 ns
```

That the operations on decimal are slower than the operations on f64 is expected and a slowdown factor of around 40x for multiplication is acceptable. The slowdown factor of around 10.000x for division is unfortunately prohibitive for my use case, calls to decimal division were very dominant in my performance profile.

**Benchmark code:**
```rust
use rust_decimal_macros::dec;
use std::time::Instant;

fn bench<T: std::fmt::Debug + Copy + std::ops::AddAssign<T>>(name: &'static str, a: T, b: T, s: T, f: impl Fn(T, T) -> T) {
  const NUM_ITERATIONS: u128 = 10000;

  // warmup
  let mut s = s;
  for _ in 0..=100 {
    s += f(a, b);
  }

  // bench
  let start = Instant::now();
  for _ in 0..NUM_ITERATIONS {
    s += f(a, b);
  }

  // print the dummy value so that the compiler can't optimize our computations away
  println!("dummy: {:?}", s);

  println!("{} : {} ns", name, start.elapsed().as_nanos() as f64 / NUM_ITERATIONS as f64);
}

fn main() {
  bench("rust-decimal division", dec!(1.02343545345), dec!(0.35454343453), dec!(0.0), |a, b| a / b);
  bench("rust-decimal multiplication", dec!(1.02343545345), dec!(0.35454343453), dec!(0.0), |a, b| a * b);
  bench("f64 division", 1.02343545345_f64, 0.35454343453_f64, 0.0_f64, |a, b| a / b);
  bench("f64 multiplication", 1.02343545345_f64, 0.35454343453_f64, 0.0_f64, |a, b| a * b);
}
```

paupino:
[...]

Issue #169, opened 2019-03-30
Support the `pow` method for `Decimal`/floating point arguments


AdminXVII:
Ion would like to remove its dependency to C, and the next step would be to remove the `decimal` crate. Unfortunately, not being able to raise some Decimal to a floating point power is a serious barrier to our objective.

paupino:
I'll take a look into how to efficiently implement this!

paupino:
Side note, it'll probably be leveraging https://docs.rs/num/0.2.0/num/pow/trait.Pow.html

ambaxter:
Adding to that - logn, log10, ln, sqrt, and nth_root would be really nice :)

paupino:
Funnily enough, in order to implement `pow` completely I need `ln` support. Consequently, this is partially in scope!

c410-f3r:
Any update regarding this matter?

paupino:
I'm afraid I've been a little side tracked lately - I'll see if I can kickstart this again since we're going into holidays and give an updated timeline.

c410-f3r:
I could do my own research and implement a high performance `pow`, `ln` or `log` implementation but it would probably take some time because of my inexperience at this field.
Having said that, can I start a bounty for this issue?

paupino:
Certainly welcome to. Implementing `pow` for integers is relatively easy, however for decimals requires implementing logarithmic tables. 

Out of curiosity, do you need this for decimal based power functions? Also, would conversion to `f64` and then back work?

AdminXVII:
Conversion to f64 could work if that makes it easier for you. The application is a scientific calculator, so while it could be useful to have exact results, f64 approximation would be ok I believe.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 91.0 TUSD (91.0 USD @ $1.0/TUSD)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/paupino/rust-decimal/169/4456).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/paupino/rust-decimal/169/4456).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://chat.gitcoin.co/'>Gitcoin Chat</a>
[...]

PR #281, opened 2020-11-02
Statistical Math functions

remkade:
Adds support for the following statistical and general math functions:
* powi
* sqrt
* exp (exponential function)
* norm_cdf (cumulative distribution of a normal function)
* norm_pdf (Normal probability density function)
* ln (natural logarithm)
* erf (error function)

I'm definitely not a math expert, so critique is welcome!

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1518,7 +1518,7 @@ impl Decimal {
     /// The estimated exponential function, e<sup>x</sup>, rounded to 8 decimal places. Stops
     /// calculating when it is within tolerance is roughly 0.000002 in order to prevent
     /// multiplication overflow.
-    fn exp(&self) -> Decimal {
+    pub fn exp(&self) -> Decimal {
         if self == &Decimal::zero() {
             return Decimal::one();
         }
@@ -1543,7 +1543,7 @@ impl Decimal {
     }
 
     /// Raise self to the given unsigned integer exponent: x<sup>y</sup>
-    fn powi(&self, exp: u64) -> Decimal {
+    pub fn powi(&self, exp: u64) -> Decimal {
         if exp == 0 {
             Decimal::one()
         } else if exp == 1 {
@@ -1558,7 +1558,7 @@ impl Decimal {
     }
 
     /// The square root of a Decimal. Uses a standard Babylonian method.
-    fn sqrt(&self) -> Option<Decimal> {
+    pub fn sqrt(&self) -> Option<Decimal> {
         if self.is_sign_negative() {
             return None;
         }
@@ -1582,7 +1582,7 @@ impl Decimal {
 
     /// The natural logarithm for a Decimal. Uses a [fast estimation algorithm](https://en.wikipedia.org/wiki/Natural_logarithm#High_precision)
     /// This is more accurate on larger numbers and less on numbers less than 1.
-    fn ln(&self) -> Decimal {
+    pub fn ln(&self) -> Decimal {
         if self.is_sign_positive() {
             if self == &Decimal::one() {
                 Decimal::zero()
@@ -1600,7 +1600,7 @@ impl Decimal {
     }
 
     /// Abramowitz Approximation of Error Function from [wikipedia](https://en.wikipedia.org/wiki/Error_function#Numerical_approximations)
-    fn erf(&self) -> Decimal {
+    pub fn erf(&self) -> Decimal {
         if self.is_sign_positive() {
             let one = &Decimal::one();
 
@@ -1619,13 +1619,13 @@ impl Decimal {
     }
 
     /// The Cumulative distribution function for a Normal distribution
-    fn norm_cdf(&self) -> Decimal {
+    pub fn norm_cdf(&self) -> Decimal {
         (Decimal::one() + (self / Decimal::from_str("1.4142135623730951").unwrap()).erf())
             / Decimal::from_str("2").unwrap()
     }
 
     /// The Probability distribution function for a Normal distribution
-    fn norm_pdf(&self) -> Decimal {
+    pub fn norm_pdf(&self) -> Decimal {
         (-(self.powi(2) / Decimal::new(2, 0))).exp()
             / (Decimal::from_str("1.4142135").unwrap()
                 * Decimal::from_str("3.14159").unwrap().sqrt().unwrap_or_default())
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1540,27 +1540,8 @@ impl Decimal {
     /// calculating when it is within tolerance is roughly 0.000002 in order to prevent
     /// multiplication overflow.
     pub fn exp(&self) -> Decimal {
-        if self == &Decimal::zero() {
-            return Decimal::one();
-        }
-
         let tolerance = Decimal::new(2, 7);
-        let mut term = self.clone();
-        let mut result = self.clone() + Decimal::one();
-        let mut prev_result = Decimal::zero();
-        let mut factorial = Decimal::one();
-        let mut n = TWO;
-
-        // Needs rounding because multiplication overflows otherwise.
-        while (result - prev_result).abs() > tolerance && n < Decimal::new(24, 0) {
-            prev_result = result;
-            term = self * term.round_dp(8);
-            factorial *= n;
-            result += (term / factorial).round_dp(8);
-            n += Decimal::one();
-        }
-
-        result
+        self.exp_with_tolerance(tolerance)
     }
 
     /// The estimated exponential function, e<sup>x</sup>, rounded to 8 decimal places. Stops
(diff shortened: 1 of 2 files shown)


remkade:
Related to: #169

src/decimal.rs
```diff
@@ -1514,6 +1514,122 @@ impl Decimal {
...
+
+    /// Raise self to the given unsigned integer exponent: x<sup>y</sup>
+    fn powi(&self, exp: u64) -> Decimal {
+        if exp == 0 {
+            Decimal::one()
+        } else if exp == 1 {
+            self.clone()
+        } else {
+            let mut result = self.clone();
+            for _ in 1..exp {
```
paupino, line 1553: There are some optimizations we can look into here potentially.

1. The easiest one is to reduce the number of calculations required. e.g. `3^9` could be broken down into `(((3^2)^2)^2)*3`. It's subtle, but instead of `9` multiplications, it would only do `4`. 
2. The second is to avoid the boxing/unboxing each loop by using an intermediary format.

Some experimentation was done a while ago on [this here](https://github.com/paupino/rust-decimal/compare/feature/pow#diff-21e0bf62805e3ff1074a826a701e3e419044270d9c8b2c9cbf4090d519c8da0bR2754).

I quite like the `powi` version though as it makes it clear and straight forward to what is required.

Anyway, probably not necessarily something to hold this up but just adding comment.

src/decimal.rs
```diff
@@ -1514,6 +1514,122 @@ impl Decimal {
...
+
+    /// The estimated exponential function, e<sup>x</sup>, rounded to 8 decimal places. Stops
+    /// calculating when it is within tolerance is roughly 0.000002 in order to prevent
+    /// multiplication overflow.
+    fn exp(&self) -> Decimal {
+        if self == &Decimal::zero() {
+            return Decimal::one();
+        }
+
+        let tolerance = Decimal::new(2, 7);
```
paupino, line 1526: Do we ever see this being something that is passed in as an argument? i.e. would it make sense to have two functions for this?

src/decimal.rs
```diff
@@ -1514,6 +1514,122 @@ impl Decimal {
...
+        let mut term = self.clone();
+        let mut result = self.clone() + Decimal::one();
+        let mut prev_result = Decimal::zero();
+        let mut factorial = Decimal::one();
+        let mut n = Decimal::new(2, 0);
+
+        // Needs rounding because multiplication overflows otherwise.
+        while (result - prev_result).abs() > tolerance && n < Decimal::new(24, 0) {
```
paupino, line 1534: This could be optimized quite a bit by preventing the boxing/unboxing. We should probably make the new function a `const fn` to hint to the compiler too.

src/decimal.rs
```diff
@@ -1514,6 +1514,122 @@ impl Decimal {
...
+    }
+
+    /// The natural logarithm for a Decimal. Uses a [fast estimation algorithm](https://en.wikipedia.org/wiki/Natural_logarithm#High_precision)
+    /// This is more accurate on larger numbers and less on numbers less than 1.
+    fn ln(&self) -> Decimal {
+        if self.is_sign_positive() {
+            if self == &Decimal::one() {
+                Decimal::zero()
+            } else {
+                let pi = Decimal::from_str("3.141592653589793238462643383279502884197169399375105820974").unwrap();
```
paupino, line 1590: This could be a useful public constant if we're deciding to define it. We could then avoid the string unwrap/parsing.

paupino (commented):
Thank you for your contribution! 

One thing I'm thinking about as I read this is that it could be optimized for performance quite a bit. There are some things where the library itself could be improved to better support, and others whereby we avoid parsing from strings etc and boxing/unboxing. 

All in all, it's adding functionality that is missing right now which is a net positive however I would like to give a go at optimizing it a bit more. I'm going to try to devote some time to getting the next version of Rust Decimal prepared over the next couple of weeks so I'll have a bit of a play around myself too (e.g. converting some things to use `const fn`).

To begin with we should _definitely_ start moving some values to constants. e.g. rather than `Decimal::new(2, 0)` or `Decimal::from_str("2")` we should have a constant such as:

```
const TWO: Decimal = Decimal {
    flags: 0,
    lo: 2,
    mid: 0,
    hi: 0,
};
```

remkade:
Thanks so much! I'll work on optimizing these.

remkade:
I've added benchmarks now for all of the stat functions. I figured this would give us some specific things to optimize.

Looks like `ln`, `norm_pdf`, and `sqrt` are by far the slowest.

```
     Running target/release/deps/lib_benches-951a632ffd1e351c

running 41 tests
test add_10k                 ... bench:      29,023 ns/iter (+/- 2,473)
test add_negative_pi         ... bench:         163 ns/iter (+/- 1)
test add_negative_point_five ... bench:          14 ns/iter (+/- 0)
test add_one                 ... bench:          18 ns/iter (+/- 0)
test add_one_hundred         ... bench:          19 ns/iter (+/- 0)
test add_pi                  ... bench:         160 ns/iter (+/- 1)
test add_point_zero_one      ... bench:           5 ns/iter (+/- 0)
test add_two                 ... bench:          18 ns/iter (+/- 0)
test decimal_from_str        ... bench:       1,502 ns/iter (+/- 36)
test div_10k                 ... bench:     240,641 ns/iter (+/- 19,824)
test div_negative_pi         ... bench:      24,402 ns/iter (+/- 250)
test div_negative_point_five ... bench:         358 ns/iter (+/- 14)
test div_one                 ... bench:         188 ns/iter (+/- 4)
test div_one_hundred         ... bench:         501 ns/iter (+/- 10)
test div_pi                  ... bench:      24,538 ns/iter (+/- 1,367)
test div_point_zero_one      ... bench:         189 ns/iter (+/- 6)
test div_two                 ... bench:         384 ns/iter (+/- 7)
test erf                     ... bench:     142,323 ns/iter (+/- 8,634)
test exp                     ... bench:     439,370 ns/iter (+/- 11,156)
test iterator_individual     ... bench:          70 ns/iter (+/- 1)
test iterator_sum            ... bench:          69 ns/iter (+/- 0)
test ln                      ... bench:   5,173,975 ns/iter (+/- 62,530)
test mul_negative_pi         ... bench:          91 ns/iter (+/- 3)
test mul_negative_point_five ... bench:           7 ns/iter (+/- 0)
test mul_one                 ... bench:           7 ns/iter (+/- 0)
test mul_one_hundred         ... bench:           7 ns/iter (+/- 0)
test mul_pi                  ... bench:          91 ns/iter (+/- 2)
test mul_point_zero_one      ... bench:           7 ns/iter (+/- 0)
test mul_two                 ... bench:           7 ns/iter (+/- 0)
test norm_cdf                ... bench:     240,720 ns/iter (+/- 6,406)
test norm_pdf                ... bench:     897,218 ns/iter (+/- 34,991)
test powi                    ... bench:         993 ns/iter (+/- 33)
test sqrt                    ... bench:   1,002,366 ns/iter (+/- 68,856)
test sub_10k                 ... bench:      32,258 ns/iter (+/- 743)
test sub_negative_pi         ... bench:         160 ns/iter (+/- 2)
test sub_negative_point_five ... bench:          13 ns/iter (+/- 0)
test sub_one                 ... bench:          19 ns/iter (+/- 0)
test sub_one_hundred         ... bench:          20 ns/iter (+/- 0)
test sub_pi                  ... bench:         161 ns/iter (+/- 2)
test sub_point_zero_one      ... bench:           7 ns/iter (+/- 0)
test sub_two                 ... bench:          19 ns/iter (+/- 1)

test result: ok. 0 passed; 0 failed; 0 ignored; 41 measured; 0 filtered out
```

paupino:
Benchmarks are great! Admittedly some may be quite tricky to speed up without [some other fundamental improvements requiring optimization](https://github.com/paupino/rust-decimal/issues/276).  
e.g. `sqrt` in it's current form, for example, will be slow due to division in a loop. `ln` uses `sqrt` so it makes sense that it's a bit slower again. 

I think for now, the low hanging fruit makes sense to get while some of the fundamental issues can be dependent on looking at #276 .

src/decimal.rs
remkade: I went ahead and updated the calculation so it works like you suggested. Looks like it makes it about 33% faster.
`654 ns/iter (+/- 23)` from 993

src/decimal.rs
remkade: I aded this as a constant, along with `TWO`. It doesn't speed up too much, but it does help.

src/decimal.rs
remkade: I could see it happening, I'll go ahead and create a version with `tolerance` as a argument. Its pretty tricky to hit a nice balance without causing multiplication to overflow though.

remkade: OK I added a version `exp_with_tolerance` and tests for it.

src/decimal.rs
remkade: I'm actually not quite sure how to properly optimize these to prevent the boxing problems. Do you have any resources on how that works? I'd love to learn in more detail and my googling hasn't been finding anything.

remkade:
Improved `norm_pdf` performance by more than half.
`897,218 ns/iter` to `367,572 ns/iter`

remkade:
I believe this is ready to merge unless you have any more things I should work on. Let me know!

PR #289, opened 2020-12-04
#271: Adds Scientific notation formatting to format strings

### LINKED ISSUE #271: feature request: method to print decimal as scientific notation
as title.

i hope there is a function to print the decimal as a string of scientific notation like `"1e-5"` from 0.00001.

paupino:
Fixes #271 

Adds scientific notation formatting via `LowerExp` and `UpperExp` traits. 

e.g. `format!("{0:e}", "42.0")` outputs `4.2e1`

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -5,6 +5,12 @@ default_to_workspace = false
 command = "cargo"
 args = ["build"]
 
+# Always test no-std with std tests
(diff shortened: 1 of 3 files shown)


paupino:
Ugh, was running `cargo make test` instead of `cargo make test-all`. Will resolve these now.

PR #296, opened 2020-12-29
Faster serde

jean-airoldie:
I had a use case where my bottleneck was due to decimal deserialization via bincode, so I though I would do some low hanging fruit optimizations.

Feedback would be much appreciated.

## Overview
* Adds arrrayvec as a dependency since it will replace the heap allocated vectors and strings
* Removes heap allocation for both serialization & deserialization with the `serde-str` feature enable. This is done my removing heap allocation for the from_str impl as well as using a to_array_str impl instead of using to_string.
* Adds a from_str_radix specialization when radix is 10, since its the most common case (~25% performance improvement)
* Improve the to_string algorithm by making it 2 pass instead of 3 pass

## Outstanding issues
* Need to add tests for the `to_array_str` method
* Im not sure the constants I used for my arrays are the right ones. Basically we wanna use the highest possible size necessary otherwise it will panic when pushing a sufficiently long decimal.

## Performance improvements according to the micro benchmarks:
```
before:
test deserialize_bincode     ... bench:       6,544 ns/iter (+/- 3,308)
test serialize_bincode       ... bench:      15,397 ns/iter (+/- 7,748)
test decimal_to_string       ... bench:       7,288 ns/iter (+/- 77)
test decimal_from_str        ... bench:       5,684 ns/iter (+/- 180)

after:
test deserialize_bincode     ... bench:       2,884 ns/iter (+/- 1,448)
test serialize_bincode       ... bench:       5,248 ns/iter (+/- 57)
test decimal_to_string       ... bench:       4,203 ns/iter (+/- 48)
test decimal_from_str        ... bench:       1,592 ns/iter (+/- 810)
```
I also tested against my real life application and the performance improvements where similar. I got from about 500k reads/s to 1.3M reads/s. I did not test the serialization path however.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ postgres = { default-features = false, optional = true, version = "0.18" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.6" }
+arrayvec = "0.5.2"
 
 [dev-dependencies]
 bincode = "1.3"
--- a/benches/lib_benches.rs
+++ b/benches/lib_benches.rs
@@ -4,6 +4,7 @@ extern crate test;
 
 use core::str::FromStr;
 use rust_decimal::Decimal;
+use bincode::Options as _;
 
 macro_rules! bench_decimal_op {
     ($name:ident, $op:tt, $y:expr) => {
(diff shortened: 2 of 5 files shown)


jean-airoldie:
Also just a general word of advise, when using a `Vec` I would recommend to always try to `Vec::with_capacity` with a over approximated capacity of what you would need. Otherwise your vec might allocate multiple times, which is expensive.

Also, you might want to checkout smallvec which is an hybrid between a vec and a arrayvec where you only allocate if you exceed the capacity. I would say its always preferable to using a vec in any performance sensitive setting, unless your allocating huge amounts of things.

paupino:
Awesome - great timing too, as I'm starting to comb through some optimizations over the next few days using a similar approach. I'll pull it once it's passing and take a deeper look into the changes. 

Btw, I'm using `cargo make` now to try to automate some of the common functions/feature combinations in the project - you can run `cargo make test-all` to run various combinations to track down where it's failing (looks like pg parsing has some rounding differences?).

jean-airoldie:
Looks like testing for postgres locally is gonna be a pain since you need to have a db running or something.

paupino:
Since I'm in there already, I can run those tests for you. Let me do some digging to see what's going on.

paupino:
Looks like you found it! I think we should add one more test to `it_can_parse_highly_significant_numbers()`, namely:

```rust
("1.234567890123456789012345678949999", "1.2345678901234567890123456789"),
```

This captures what was happening in pg.

jean-airoldie:
Alright so i think I got it. So looks like the midpoint calculation were always wrong. I'll make it so the tests pass but I'll put some FIXME comments.

jean-airoldie:
I'm not sure what the implications are so i'll keep it how it was.

jean-airoldie:
But I'm guessing that maybe the rounding was wrong all along and the tests reflect that?

paupino:
Hmm, I'm not sure that's the fix, but agree that something is a bit weird. My suspicion is that it's actually the fixed size array that is interfering with `coeff.len()` calcs. Interestingly, you can get them passing by changing the max precision from `29` to `30` however I want to understand what is triggering the issue if possible. I'll run the code a bit more: I can see a few areas for optimization so will take a stab at cleaning that up while I'm there (e.g. `a-Z` parsing isn't required for radix 10)

jean-airoldie:
I got the same error from fixing the midpoint on the master branch directly (which doesn't use arrays) so I'm relatively sure that its the cause.

jean-airoldie:
As for the the fix size array, it should behave like the vector expect it doesn't have all the methods that vec has and should panic if the capacity overflows. So if it was a issue caused by the arrays, they would panic.

jean-airoldie:
Check this PR https://github.com/paupino/rust-decimal/pull/299 that shows that the issue is present on master.

paupino:
Oh, I see what you mean. Cool, in that case it's not the midpoint stuff that needs changing (i.e. to 10 etc) - it's the estimated max size which is wrong. For `10` this should be 28. I'll need to recalculate these to just double check it all.

paupino:
Cool; after playing around with it I can see some clear improvements. Made some small refactors when I was reviewing so did a ff-merge to get the PR in. Thank you for doing this - two clear wins in my opinion! (midpoint logic issue + speed improvements)

PR #299, opened 2020-12-30
Replicate the midpoint bug on master

jean-airoldie:
Here is a demonstration of the midpoint bug, for sanity check purposes.


paupino:
Closing, merged test and fix as part of #296.

PR #302, opened 2020-12-31
feat: tokio 1.0

### LINKED ISSUE #293: trait FromSql is not implemented for rust_decimal::Decimal with postgres 0.19
When using the 0.19 version of the postgres library, I get the error above. When I changed to version 0.18 it works without any issues.

gakonst:
Tokio 1.0 & postgres updates, should close #293 @paupino


paupino:
Cool; at some stage I'm going to need to figure out a better way to maintain these dependencies. For now, however, this should work just fine.

PR #306, opened 2021-01-09
Fixes subtraction overflow by wrapping

### LINKED ISSUE #305: Division underflow on master
I found a way to make division panic while doing some unrelated prop-testing.

Dividing "32.91625929004387114334488" by "3.27629537"  causes a "attempt to subtract with overflow" panic.

paupino:
Fixes #305


paupino:
Moved around the features a bit so that no-std also uses the new algorithms. It can be explicitly disabled instead.

PR #307, opened 2021-01-11
Add TryFrom for f32/f64 to/from Decimal

### LINKED ISSUE #265: implement std conversion traits for f32/f64
as in the Rust API guidelines: https://rust-lang.github.io/api-guidelines/interoperability.html#conversions-use-the-standard-traits-from-asref-asmut-c-conv-traits

having to look up the crate docs to discover the special-sauce "FromPrimitive" trait i need to use to get `from_f64` instead of using `.try_into()` is annoying.

okaneco:
Implement TryFrom\<f32/f64\> for Decimal
Implement TryFrom\<Decimal\> for f32/f64
Add tests for try_into f64, try_from f32/f64
Fix some clippy lints
- Remove unnecessary returns
- Remove unnecessary clone on Copy type
- Simplify if-chain that returned bool instead of the conditional

This should close #265

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -654,7 +654,7 @@ impl Decimal {
     /// ```
     pub fn max(self, other: Decimal) -> Decimal {
         if self < other {
-            return other;
+            other
         } else {
             self
         }
@@ -671,7 +671,7 @@ impl Decimal {
     /// ```
     pub fn min(self, other: Decimal) -> Decimal {
         if self > other {
-            return other;
+            other
         } else {
             self
         }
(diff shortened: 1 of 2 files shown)


src/decimal.rs
```diff
@@ -3958,6 +3954,38 @@ impl ToPrimitive for Decimal {
     }
 }
 
+impl core::convert::TryFrom<f32> for Decimal {
+    type Error = crate::Error;
+
+    fn try_from(value: f32) -> Result<Self, Error> {
+        Self::from_f32(value).ok_or_else(|| Error::new("Failed to convert to Decimal"))
```
okaneco, line 3961: I wasn't sure what message these errors should contain. Since From/ToPrimitive return Option, there's no information about what failed in particular.

paupino:
This is great! Thank you.

PR #311, opened 2021-01-20
Fix no-std environment

c410-f3r:
The default features of `arrayvec` assume a std environment


paupino:
Good catch. I'm going to need to revisit the CI tests to see why this wasn't caught with the `no-std` tests.

PR #312, opened 2021-01-23
Fix a logical error in Neg implementation

KonishchevDmitry:
Negation is broken in 1.10.0.

#303 introduces a logical error into `Neg` implementation - it always makes the value negative instead of reverting the sign:

```rust
let quantity = dec!(-11);
dbg!(quantity);
dbg!(-quantity);
```

```
[test.rs:12] quantity = -11
[test.rs:13] -quantity = -11
```


paupino:
Thank you. I'll release a hotfix containing this shortly.

PR #350, opened 2021-04-23
Avoid Error::new in from_scientific except in error cases

thomcc:
Error::new allocates a String (via str's Into<String> impl), so there's no reason to pay this cost in the success case.


paupino:
Awesome, thanks for fixing this @thomcc

PR #351, opened 2021-04-23
Version 1.12 with some internal housekeeping

paupino:


abreis:
@paupino The removal of `Zero` from the prelude is a semver breaking change 🙂 this broke our pipelines.

paupino:
Ugh, sorry - I'll add that back to `prelude`.

paupino:
Hopefully that should now be fixed now @abreis. I yanked `1.12.0` and added `1.12.1` which adds `Zero` back to prelude.

abreis:
Thanks so much for the quick response!

PR #356, opened 2021-04-24
Fix `0 - 0 = -0` -> `0 - 0 = 0`

KonishchevDmitry:
Fixes a little regression introduced in 1.12.


paupino:
Awesome, thank you. Captured so many other cases and missed the most fundamental case!

Issue #354, opened 2021-04-23
Version 2 Roadmap planning


paupino:
Some thoughts for version 2 of the Decimal library. Issue numbers will be added as they are created.

* [ ] Define updated rust minimum version.
* [ ] Remove `deprecated` functions.
* [ ] Review API design. It is a little inconsistent in places - it'd be nice to make this more consistent and extensible. Some guidelines for API design may be useful. Related to #49.
* [ ] Investigate `u128` optimizations when performing 96 bit operations. The hypothesis is that I can squeeze further performance from more complex numbers. Related to #135.
* [ ] Investigate variable precision decimals, perhaps leveraging constant generics. _Reduced_ precision is easy, however greater precision is an interesting challenge since current optimizations won't hold true. This requires some experimentation to figure out if anything makes sense within this library or whether it's a variation. Related to #135 and #310.

cereum:
Returning a Result<Decimal, ErrorEnum>  would make using the checked operations much nicer. When I use checked math operations I would like to be able to unwrap them with ?, this way I can make a top level error handler for math errors vs a match statement for each checked operation. Additionally it would make my code much more readable.

paupino:
Brainstorming thread:

Lazily evaluated rational numbers? i.e. maintain precision until absolutely necessary. Could be a different type using decimal as a subset.
Const generics - should this be the underlying type (e.g. u32 vs u64 vs u128) or the size of the words (e.g. 4, 8, 16 etc).

zzhengzhuo:
Sorry to bother you.
I think the current version is stable enough. Would you consider starting the upgrade to v2?

paupino:
Hi @zzhengzhuo - sorry for the delay in reply. Truth be told I haven't had much time to focus on this project (or others for that matter) as of late. But I do agree that we could probably, at the very least, start a branch revisiting the API/Errors and cleaning that up. From there we could really just focus on the internals (if need be).
If you're willing to start a new PR/development I'd be more than happy to branch off v1 so that development of v2 breaking changes can be performed on the main branch. I really appreciate the offer for assistance!

zzhengzhuo:
Thanks for your response. I'll be starting development on the master branch soon. I'll likely begin with feature #310, as it seems to involve significant breaking changes.

Tony-Samuels:
Do you know what features you want included in 2.0, @paupino? I wasn't sure if #310 would make the cut. It's on the list in the original message, but only as an investigation.

[...]

PR #360, opened 2021-04-28
Add `from_i128_with_scale_rslt` 

c410-f3r:
@paupino Feel free to close this PR if the changes aren't desirable.

My use-case: External stream of bytes that represent `i128` decimals are received but can't be trusted so a check is necessary to ensure correctness and recovery. With `from_i128_with_scale`, the binary will simply halt execution as there is currently no way to verify if `i128` or `u32` are valid, thus the creation of `from_i128_with_scale_rslt`.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -218,7 +218,6 @@ impl Decimal {
     /// let max = Decimal::from_i128_with_scale_rslt(i128::MAX, u32::MAX);
     /// assert!(max.is_err());
     /// ```
-    #[must_use]
     pub fn from_i128_with_scale_rslt(num: i128, scale: u32) -> crate::Result<Decimal> {
         if scale > MAX_PRECISION {
             return Err(Error::ScaleExceedsMaximumPrecision(scale));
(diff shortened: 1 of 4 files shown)


paupino:
This PR is great - it's also starting to clean up the error handling which is long overdue. The one hangup I have is on the naming of the function. In [version 2](#354) I'll do a holistic clean up of the API, but for now I'm just wondering if there is a better name? Admittedly I had to look up what `rslt` stood for - which seems obvious now on retrospect! 
I don't have a better name just yet - perhaps this one is ok? Anyway, let me mull on it for a bit.

c410-f3r:
@paupino Thanks! Some say that naming is one of the most difficult things in the programming field :)
What the standard library do is prefixing with `try_` (`try_from` or `try_reserve`), so `from_i128_with_scale_rslt` would become `try_from_i128_with_scale`. There are also `checked_` variants like `checked_add` or `checked_mul`. In any case, feel free to pick the most appropriated name.

paupino:
I liked the `try_` approach so renamed it to that before merge - thanks for your help!

c410-f3r:
Thank you ! :D

Issue #359, opened 2021-04-27
Use a stable repr to enable safe casting


jean-airoldie:
I've been looking in storing `Decimal` inline in c struct so that I don't have to pay serialization & deserialization cost. However I've noticed that the `Decimal` type does not have a stable ABI represenation.

Would you consider using `#[repr(C)]` or `#[repr(u128])` in order to provide a stable ABI? This would basically deprecate the `serialize` & `deserialize` methods, but lock-in the current inner represation. I'm guessing this shouldn't impact performance.

For my use case, the main benefit would be to be able to safely cast a `&[u8]` into `&[Decimal]`, which currently cannot be done safely without allocating an intermediary buffer etc.

paupino:
This is actually something I looked at a while ago and would be open to adding it again. Perhaps to begin with, placing it behind a feature flag could be useful for experimentation purposes.

c410-f3r:
Just for the record. https://github.com/rust-lang/rust/issues/81996

jean-airoldie:
From my understandting of the issue mentioned by @c410-f3r, `repr(C)` is unsounds for some zero-sized types and can allow overflowing explicit discriminants for enums. So as far as I understand it, it shouldn't affect the decimal's repr.

However, for some additional sanity, I think it would be better to add some layout assertion tests, to make sure the layout is really what we expect. In that case, simply making sure the size of the decimal type would be enought, I believe.

PR #364, opened 2021-04-29
Add c-repr feature

jean-airoldie:
This adds the c-repr feature which forces `Decimal` to be #[repr(C)].

Closes https://github.com/paupino/rust-decimal/issues/359.


jean-airoldie:
I ran the benchmarks on my machine to compare with the default `repr(rust)` and found no noticeable difference in performance. I don't really know the specifics of the current rust representation, but i'm guessing that since each field is already 32bits aligned, it is the same as the `repr(C)`.

jean-airoldie:
I added a basic layout test as well.

PR #366, opened 2021-05-01
Export num_traits::One from prelude

spearman:
Fixes a breaking change introduced in 1075fff


paupino:
Cool; thank you. I'll get this merged/released shortly.

paupino:
This is actually merged and released. I had to do a little bit of a detached release to avoid getting the next version released early hence the workaround.

PR #371, opened 2021-05-04
Fix Decimal::MAX comment

falsetru:


paupino:
Good catch, thank you!

PR #374, opened 2021-05-13
Add checked_exp.

schungx:
checked_exp will fail for > 11.7578 (empirically found).

### DIFF
--- a/src/maths.rs
+++ b/src/maths.rs
@@ -54,6 +54,10 @@ pub trait MathematicalOps {
     /// sooner at the potential cost of a slightly less accurate result.
     fn exp_with_tolerance(&self, tolerance: Decimal) -> Decimal;
 
+    /// The estimated exponential function, e<sup>x</sup>. Stops calculating when it is within
+    /// tolerance of roughly `0.0000002`. Returning `None` on overflow.
+    fn checked_exp(&self) -> Option<Decimal>
+    
     /// Raise self to the given integer exponent: x<sup>y</sup>
     fn powi(&self, exp: i64) -> Decimal;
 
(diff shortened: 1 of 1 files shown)


src/maths.rs
```diff
@@ -123,6 +127,15 @@ impl MathematicalOps for Decimal {
 
         result
     }
+    
+    fn checked_exp(&self) -> Option<Decimal> {
+        // Current calculation method will overflow when e is raised to a power larger than around 11.7578.
+        if self > Decimal::from_parts(117578, 0, 0, false, 4) {
```
paupino, line 133: Hmm, I'm not sure if this is the approach I'd go. While it's a nice sanity check, the code as you currently have it still has risk of `panic`ing which is not what I'd expect for a `checked_` function. 
It'd be more efficient and effective to modify `exp_with_tolerance` to have a checked variant. I suspect overflow either occurs on one of the lines:
```rust
term = self * term;
let next = result + (term / factorial);
```
My gut, is that it's on the multiplication. i.e. if you modify the code to use `checked_mul` and handle the `Option` result. You could in fact use the internal `ops` function if you chose to, handling the `Result` directly.

The other thing I'd request is a test for this function. Something to show it working with `Some` value and also safely handling an overflow by returning `None`.

paupino (changes requested):
Thanks for your contribution! I've added a comment requesting some changes - let me know if that makes sense.

src/maths.rs
schungx: That's right on.  The overflow is on the multiply.

schungx: I'll try your suggestion.

schungx:
I'll close this PR and open another one with the revised changes.

Issue #376, opened 2021-05-21
attempt to subtract with overflow


blasrodri:
Hi! After doing some fuzzing, I got:

```rust
use rust_decimal_macros::dec;
fn main() {
        dec!(590.327485400400946775425512) / dec!(53.68997202826239); // works fine
        dec!(590.3274854004009467754255123) / dec!(53.68997202826239); // panic
}
```

Is this a  bug? Or is there a way to prevent this panic from happening at all?

Thanks!

PR #377, opened 2021-05-21
div: update the check to exclude equality

blasrodri:
Fixes: #376


src/ops/div.rs
```diff
@@ -100,7 +100,7 @@ impl Buf16 {
         // Do a simple check to see if the hi portion of the dividend is greater than the hi
         // portion of the divisor.
         let divisor_hi32 = (divisor >> 32) as u32;
-        if self.data[2] >= divisor_hi32 {
+        if self.data[2] > divisor_hi32 {
```
paupino, line 103: This should be resolved by wrapping sub/add. The reason being is we're also assuming it can go "negative" below. The "catch" however is that we're always dealing with `u64` which of course can't go negative - so we detect this with `wrapping` functions which cause the resulting number to be < the original number.

Consequently, if you change line 107 to: `low64 = low64.wrapping_sub(divisor << 32).wrapping_add(divisor);` it'll also fix the "overflowing sub" issue.

Rather than having you make the change to get it merged, I'll add this to your PR now and get it merged into master.

paupino (commented):
Thank you for identifying this issue and suggesting a fix! I've made a comment as to what I think is the correct fix for this panic but will instead just merge that change in now (since it's one line!).

src/ops/div.rs
blasrodri: Thanks @paupino ! I really appreciate you taking all the effort and time to support this crate so well.

paupino: For reference, the changes I made before merge were this: https://github.com/paupino/rust-decimal/commit/0acadfe68c58453d42a12b14f403d40a60d34d43

PR #382, opened 2021-05-24
Protect against exp underflow.

schungx:


paupino:
I've just been looking into this - I wonder if the better approach for negative `exp` is to instead rewrite the equation to: `1 / e^n`?

This means that for the `-8` example we should be able to rewrite it to the equivalent of:

```rust
dec!(1) / dec!(8).exp()
```

Of course, this trends towards zero effectively underflowing, however returning zero in this case is probably ok. In fact, simply letting the division take care of the underflow should be enough.

One of the advantages with this is that I think it'll be more accurate for negative numbers than it is currently. As an example, `1/e^8` evaluates to ~`0.00033546` whereas `e^-8` evaluates to ~`0.0002858169660624369145768176`. Wolfram Alpha for this example gives `0.0003354626279025118388213891257808610193109001337203193605445757`.

paupino:
I've actually just recreated this PR as I wanted to evaluate the hypothesis - it does appear to be _more_ accurate for negative numbers which looks good. Perhaps I'll close this one and you can help review the other? #385

schungx:
Great!  Closing this one.

Issue #380, opened 2021-05-23
Write a guide for how to build/test from source


paupino:
This should help first time contributors as well as explain some of the recommended build tooling to use for testing.

PR #383, opened 2021-05-24
Add fuzz targets to exercize checked functions

c410-f3r:
Useful to catch user inputs that could abort thread execution like #376.

* Adds limited CI execution through a script that can also be used for local development.
* Constify some methods
* Creates a new `try_new` method for convenience.
* Adds initial `constructors`, `add_and_sub` and `div_and_mul` targets (more yet to come).

CI is purposely failing because things like:
```rust
let a = X;
let b = Y;

let c = a / b;

assert_eq!(c * b, a);
```
Are currently being used, which brings rounding errors like:
```
left: `127.99999999986666197612373336`
right: `128`
```
If it is a normal behaviour (probably is), then the above code snippet can be replaced by:
```rust
let a = X;
let b = Y;

let _c = a / b;
let _d = a * b;
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -158,6 +158,8 @@ jobs:
           profile: minimal
           toolchain: nightly
 
+      - uses: davidB/rust-cargo-make@v1
+
       - uses: actions-rs/install@v0.1
         with:
           crate: cargo-fuzz
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -165,7 +165,7 @@ jobs:
           crate: cargo-fuzz
           use-tool-cache: true
 
-      - name: Run macro tests
+      - name: Run fuzz tests
         uses: actions-rs/cargo@v1
         with:
           command: make
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,6 +144,31 @@ jobs:
           command: clippy
           args: --workspace --all-features
 
+  fuzz:
+    name: Fuzz
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          override: true
+          profile: minimal
+          toolchain: nightly
+
+      - uses: actions-rs/install@v0.1
+        with:
+          crate: cargo-fuzz
+          use-tool-cache: true
+
+      - name: Run macro tests
+        uses: actions-rs/cargo@v1
+        with:
+          command: make
+          args: fuzz
+
   minimum_rust_version:
     name: Check minimum rust version
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,8 @@ Cargo.lock
 
 # IDE support
 .idea/
+
+# Fuzz
+artifacts
+corpus
+target
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -1,6 +1,9 @@
 [config]
 default_to_workspace = false
 
+[env]
+FUZZ_RUNS = 10000000
+
 [tasks.build]
 command = "cargo"
 args = ["build"]
@@ -70,6 +73,24 @@ workspace = true
 install_crate = "rustfmt"
 command = "cargo"
 args = ["fmt", "--", "--emit=files"]
+
+[tasks.fuzz]
+dependencies = [
+    "fuzz-arithmetic",
+    "fuzz-constructors"
+]
+
+[tasks.fuzz-arithmetic]
+toolchain = "nightly"
+install_crate = "cargo-fuzz"
+command = "cargo"
+args = ["fuzz", "run", "arithmetic", "--", "-runs=${FUZZ_RUNS}"]
+
+[tasks.fuzz-constructors]
+toolchain = "nightly"
+install_crate = "cargo-fuzz"
+command = "cargo"
+args = ["fuzz", "run", "constructors", "--", "-runs=${FUZZ_RUNS}"]
 
 [tasks.outdated]
 install_crate = "cargo-outdated"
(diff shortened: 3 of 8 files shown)


paupino:
Thanks for putting this together. It appears that it's failing on an inconsistency in error messaging due to the change here:

https://github.com/paupino/rust-decimal/pull/383/files#diff-21e0bf62805e3ff1074a826a701e3e419044270d9c8b2c9cbf4090d519c8da0bR181

There are two ways to address this. The first is by adding the `the` word to this: https://github.com/paupino/rust-decimal/blob/master/src/error.rs#L38

The other way is to change the test to ignore the `the` word:
https://github.com/paupino/rust-decimal/blob/master/tests/decimal_tests.rs#L2985

I think I prefer fixing `error.rs` as `Scale exceeds the maximum precision allowed: 29 > 28` sounds better to `Scale exceeds maximum precision allowed: 29 > 28` to me?

paupino:
Btw, I'm more than happy to fix it for you as part of a merge but wanted to make sure you were ready first!

c410-f3r:
Sup @paupino
Thanks for the suggestion, just applied the suggested fix. Rounding errors from the fuzz targets are now the only thing remaining.

paupino:
I've fixed the negating overflow that this discovered (awesome!) so am just rerunning the tests now. Also, to answer your questions re:

```rust
let a = X;
let b = Y;

let c = a / b;

assert_eq!(c * b, a);
```

I'm not sure this is possible due to loss of precision for some rational numbers. e.g. `1 / 3` of course equates to `0.3` with `3` recurring. Due to the storage format we need to effectively "underflow" and stop calculating the recurring digit after a while. Because we're not storing the rational representation, when we multiply it back it'll be `0.9` with `9` recurring.

paupino:
The current fuzz is failing due to a similar problem. In mathematics we can assume the following:

```rust
let c = a + b;
assert_eq!(c - b, a);
assert_eq!(c - a, b);
```

In practice, however, it's a little more convoluted. If you assume we have the following:

```rust
let a = dec!(-77738765904499.114450305941545);
let b = dec!(0.0000001082331824127);
```

In this case when we do `a + b` we get `c` equalling: `-77738765904499.114450197708363`. It's subtle, but in this we've actually _lost_ some precision due to losing space to store it in the Decimal type. If we had unlimited precision we'd end up with: `-77738765904499.1144501977083625873`.

Because of this underflow handling (i.e. rounding) it means that we're unable to assert that `c - a` (in this case) is equal to `b` since we've lost precision. This is expected behaviour for fixed precision decimal types - it all depends on the maximum precision as to where it starts underflowing. (nb: in version 2 of the library we'll take a look at the possibility of configurable precision decimals using const generics).

c410-f3r:
Thank you very much for all the information, @paupino It is then more like a feature than a bug. My apologies for the noise.

To keep CI green, arithmetic tests were commented but I will try to fix (on my free-time) some underlying issues to uncomment them all.

> (nb: in version 2 of the library we'll take a look at the possibility of configurable precision decimals using const generics).
Really excited for this feature!

scripts/fuzz.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# Fuzzing tests used by CI. Free feel to modify or remove `RUNS` for local development or local tests.
+
+set -euxo pipefail
+
+RUNS=1000000
```
paupino, line 7: Since the project and github actions are both using `cargo-make` you could actually implement this there for cross platform compatibility, self installing cargo extensions etc. To do this:

To the top of `Makefile.toml`:
```toml
[env]
FUZZ_RUNS = 1000000
```

In the tasks section of `Makefile.toml`:
```toml
[tasks.fuzz]
dependencies = [
    "fuzz-arithmetic",
    "fuzz-constructors"
]

[tasks.fuzz-arithmetic]
toolchain = "nightly"
install_crate = "cargo-fuzz"
command = "cargo"
args = ["fuzz", "run", "arithmetic", "--", "-runs=$FUZZ_RUNS"]

[tasks.fuzz-constructors]
toolchain = "nightly"
install_crate = "cargo-fuzz"
command = "cargo"
args = ["fuzz", "run", "constructors", "--", "-runs=$FUZZ_RUNS"]
```

You _should_ be able to just run `makers fuzz` to run everything, or if you wanted to isolate to just a portion of tests you'd run `makers fuzz-constructors`. 

It's where I've added all the test configurations, benchmarking logic as well as code coverage rules - so it'd be nice to include fuzz in there too.

To finish off - this is why I need to write a `BUILDING.md` document or whatever. Another vote for #380 !

c410-f3r:
Sorry for taking too long to respond, too little free-time.
It was actually easy to add these targets because you already provided everything so thank you!

.github/workflows/main.yml
```diff
@@ -144,6 +144,33 @@ jobs:
...
+          toolchain: nightly
+
+      - uses: davidB/rust-cargo-make@v1
+
+      - uses: actions-rs/install@v0.1
+        with:
+          crate: cargo-fuzz
+          use-tool-cache: true
+
+      - name: Run macro tests
```
paupino, line 168:
```suggestion
      - name: Run fuzz tests
```

paupino:
I think this is ready for merge now. I'm going to do a quick hotfix release before merging it in.

Also, a note to self: I'll need to document this in `BUILDING.md` also, as well as what to do when it fails in the PR template.

c410-f3r:
Hey @paupino 
Sorry for the lack of update but thanks for merging these changes. Hope they will be useful for the project as a whole.

paupino:
@c410-f3r No need to apologize - I really appreciate and am truly grateful for your contribution and help! This feature has already found a couple of subtle bugs so it is already demonstrating its value! I'm looking forward to continuing to evolve this feature so that we can increase test coverage on the library as a whole.

c410-f3r:
@paupino Really appreciate your comprehension and thank you very much for all the hardwork you put into this awesome crate

Issue #398, opened 2021-06-18
`ln` panics on `Decimal::ZERO`


SebRollen:
MWE:
```rust
use rust_decimal::prelude::*;

fn main() {
    Decimal::ZERO.ln();
}
```
```shell
thread 'main' panicked at 'Division by zero', /Users/rollen/.cargo/registry/src/github.com-1ecc6299db9ec823/rust_decimal-1.14.2/src/decimal.rs:1972:45
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Of course taking the log of zero should yield some sort of error, but so should taking the log of negative numbers most likely, and for those the result is set to zero.
```rust
Decimal::new(-1, 0).ln() == Decimal::ZERO //true
```

The easiest fix might be to extend the predicate on this line to also check for `*self != Decimal::ZERO`
https://github.com/paupino/rust-decimal/blob/e346c59b352e98d4a6ca83b20aae6f38b59ce524/src/maths.rs#L328. Happy to open a PR for that

PR #399, opened 2021-06-18
fix panic in call to Decimal::ZERO.ln()

SebRollen:
Fixes https://github.com/paupino/rust-decimal/issues/398

I wasn't sure whether to write the `VERSION.md` update myself or not - of course feel free to update the language as you see fit.


paupino:
Cool, thank you for this. It may be worth us thinking long term about a `checked_ln` function with `<= 0` panicing, however since it's a breaking change we'll park that and maybe look at introducing this in a future major version.

PR #404, opened 2021-07-15
Support serializing to arbitrary precision numbers

### LINKED ISSUE #402: Roundtrip using "serde-float" and "serde-arbitrary-precision" fails
The assertion in the following code fails, whereas I would expect it to succeed based on the features I have activated:

`main.rs`:
```rust
use rust_decimal::Decimal;

fn main() {
    let dec = Decimal::new(481, 2);
    let serialized = serde_json::to_value(dec).unwrap();
    let dec2: Decimal = serde_json::from_value(serialized).unwrap();

    assert_eq!(dec, dec2);
}
```

`Cargo.toml`:
```toml
[package]
name = "rust_decimal_test"
version = "0.1.0"
authors = ["..."]
edition = "2018"

[dependencies]
rust_decimal = { version = "1.14.3", features = ["serde-float", "serde-arbitrary-precision"]}
serde_json = { version = "1.0.64", features = ["arbitrary_precision"] }
```

Output:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `4.81`,
 right: `4.8100000000000005`', src/main.rs:8:5
```

Reading the documentation of `serde-arbitrary-precision` further, it makes it clear that it only supports arbitrary precision when *parsing* decimals but not when serializing them. `.to_f64()` is still used in the `Serialize` impl which causes the rounding issue.

Would you be opposed to extending the behaviour of the `serde-arbitrary-precision` feature to also serialize `Decimal`s to JSON numbers exactly, so that the above test would pass?

I'd be willing to attempt to contribute this change if you thought it was a good idea.

JamesHinshelwood:
Closes #402

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -178,7 +178,7 @@ impl serde::Serialize for Decimal {
     }
 }
 
-#[cfg(feature = "serde-float")]
+#[cfg(all(feature = "serde-float", not(feature = "serde-arbitrary-precision")))]
 impl serde::Serialize for Decimal {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
@@ -186,6 +186,18 @@ impl serde::Serialize for Decimal {
     {
         use num_traits::ToPrimitive;
         serializer.serialize_f64(self.to_f64().unwrap())
+    }
+}
+
+#[cfg(all(feature = "serde-float", feature = "serde-arbitrary-precision"))]
+impl serde::Serialize for Decimal {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serde_json::Number::from_str(&self.to_string())
+            .map_err(serde::ser::Error::custom)?
+            .serialize(serializer)
     }
 }
 
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -276,6 +276,32 @@ mod test {
     }
 
     #[test]
+    #[cfg(all(feature = "serde-float", feature = "serde-arbitrary-precision"))]
+    fn serialize_decimal_roundtrip() {
+        let record = Record {
+            // 4.81 is intentionally chosen as it is unrepresentable as a floating point number, meaning this test
+            // would fail if the `serde-arbitrary-precision` was not activated.
+            amount: Decimal::new(481, 2),
+        };
+        let serialized = serde_json::to_string(&record).unwrap();
+        assert_eq!("{\"amount\":4.81}", serialized);
+        let deserialized: Record = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(record.amount, deserialized.amount);
+    }
+
+    #[test]
+    #[cfg(all(feature = "serde-str", not(feature = "serde-float")))]
+    fn serialize_decimal_roundtrip() {
+        let record = Record {
+            amount: Decimal::new(481, 2),
+        };
+        let serialized = serde_json::to_string(&record).unwrap();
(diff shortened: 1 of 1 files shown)


paupino:
In lieu of the comment better understanding the issue - it'd be great if you could also add a test or two which demonstrate the issue (and it being fixed).

I'm yet to add a `BUILDING.md` doc to describe how to test, but effectively you can test combinations of features using [cargo-make](https://github.com/sagiegurari/cargo-make). (e.g. `makers test-serde`). 

There is a test in there already that tests both `serde-arbitrary-precision` and `serde-float` enabled [here](https://github.com/paupino/rust-decimal/blob/b7cf68ed014aed96807ac5b76bb6d5bc5c1cdca3/Makefile.toml#L189-L191).
I'd expect this test to fail before your change and to succeed afterwards.

Thanks for your help!

JamesHinshelwood:
> There is a test in there already that tests both serde-arbitrary-precision and serde-float enabled here.

Thanks for the pointer! I will add a test or two.

JamesHinshelwood:
@paupino I've added a couple of tests for roundtripping a decimal with both the `serde-float` and `serde-str` features enabled (though I might have got the exact combinations of features wrong in the tests, let me know if you think so).

paupino:
I think that something you mentioned in the issue write up was extremely enlightening. 

> In fact, in this case I think there's no need for the arbitrary_precision features either, since Decimals can be serialized and deserialized to Strings exactly already.

That made me think... why do we bother converting to a float at all during serialization with `serde-float`? The solution that you've provided I think could be used to _replace_ the serialization for `serde-float` - i.e. 

```rust
#[cfg(feature = "serde-float")]
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
            .serialize(serializer)    
    }
}
```

You'd also want to update `cargo.toml` in this case to something like:

```
serde-float = ["serde", "serde_json"]
```

`serde-arbitrary-precision` is still a relevant feature for ensuring deserialization is accurate, however isn't _really_ required for serialization. 

The only reason I can think of for _not_ doing this is if you wanted to serialize to a float but not use `serde_json` (i.e. an alternative format like `bincode`). Which _could_ be a valid reason perhaps?

Thoughts?

paupino:
I'd also be open to keeping it as is too for now. I think `serde-float` should be reimplemented at some stage anyway, likely leveraging some of the logic that enables the arbitrary precision feature but without a dependency on `serde-json`. i.e.

```rust
        let mut s = serializer.serialize_struct(TOKEN, 1)?;
        s.serialize_field(TOKEN, &self.n)?;
        s.end()
```

In fact - I'm going to keep this as is. I'll create a ticket for optimizing `serde-float` and clean up the docs to clarify the feature combinations (they're getting a bit unwieldy!).

JamesHinshelwood:
Thanks for merging @paupino 

> That made me think... why do we bother converting to a float at all during serialization with serde-float? The solution that you've provided I think could be used to replace the serialization for serde-float - i.e.

Yes I agree. I think the `arbitrary_precision` feature could be removed. `serde_json` will delegate to either `.serialize_{u,i,f}64` or `.serialize_struct(TOKEN,1)` etc. depending on the presence of its own `arbitrary_precision` feature.

> The only reason I can think of for not doing this is if you wanted to serialize to a float but not use serde_json (i.e. an alternative format like bincode). Which could be a valid reason perhaps?

That's a good point. It also would force a dependency on `serde_json` even for non-JSON users of the library.

JamesHinshelwood:
@paupino Do you have a policy of when it would make sense to cut this into a new crates.io release? Its not an immediate issue for me since I can use a `git` dependency but it would be nice to have a proper version at some point.

PR #406, opened 2021-07-19
Fix negative zero serialization bug

edwardycl:
When negative zero (`-Decimal::ZERO`) is serialized, it becomes `"-"`

This commit fixes that bug

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -266,6 +266,14 @@ mod test {
     }
 
     #[test]
+    #[cfg(not(feature = "serde-float"))]
+    fn serialize_negative_zero() {
+        let record = Record { amount: -Decimal::ZERO };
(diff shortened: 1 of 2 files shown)


paupino (changes requested):
Some minor suggestions below, however one request I would make is for a test added showing this use case being fixed. Perhaps since this is isolated to serialization you could add it as part of the `serde.rs` tests?

Thank you for your help!

edwardycl:
Suggestions applied and a test added!

paupino (approved):
Awesome, have confirmed through the test that existing code generated:

```json
{"amount":"-"}
```

After the fix, `-0` correctly generates as:

```json
{"amount":"-0"}
```

🚀

PR #412, opened 2021-08-03
Enable all fuzz tests

c410-f3r:
After running for a while, no panic occurred so it seems that `let mut result = self + Decimal::ONE;` was the sole culprit.
It is worth noting that although no panic occurred on my local tests and no panic can occur after running the fuzz targets for many years, all explicit arithmetic operator (`+`, `*`, `-`, `%`, etc) should be audited to catch any possible corner-case.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -2,7 +2,7 @@
 default_to_workspace = false
 
 [env]
-FUZZ_RUNS = 10000000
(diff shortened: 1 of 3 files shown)


paupino:
Good catch re the +1. At some stage it may be worth breaking out the fuzz tests so that we can be more granular when they run. For now though, this is great!

PR #415, opened 2021-08-14
Add benchmarks to compare different structures 

c410-f3r:
Useful to see how `rust-decimal` is performing compared to other libraries or how much is lagging behind `f32` or `f64` hardware based operations.

addition
division
subtration
multiplication

As seen from above, `rust-decimal` is the fastest decimal library with operations ~2000% slower if compared to native primitives.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,16 +1,21 @@
+[[bench]]
+harness = false
+name = "libraries"
+path = "benches/libraries.rs"
+
 [package]
 authors = ["Paul Mason <paul@form1.co.nz>"]
 categories = ["science","data-structures"]
 description = "A Decimal Implementation written in pure Rust suitable for financial calculations."
 documentation = "https://docs.rs/rust_decimal/"
 edition = "2018"
+exclude = [ "tests/generated/*" ]
 keywords = ["decimal","financial","fixed","precision"]
 license = "MIT"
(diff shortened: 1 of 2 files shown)


paupino:
This is cool to see. Also cool to see that criterion and cargo-bench can exist together just fine.

c410-f3r:
Hum... I was thinking about removing comparisons between other rust libraries to prevent ppl pointing fingers and reduce the amount of development dependencies. So the only thing that would remain is the comparison between `rust-decimal` and native floating primitives

paupino:
Oh yeah - fair call. Perhaps a separate repo could compare all libraries instead - it is quite useful to keep track of over time but perhaps not the responsibility of this library/repo.

c410-f3r:
> Oh yeah - fair call. Perhaps a separate repo could compare all libraries instead - it is quite useful to keep track of over time but perhaps not the responsibility of this library/repo.

Indeed! Another repository would be the best approach so feel free to create one if desired

paupino:
I reverted this PR but more than happy to include again for just `f32`/`f64`! One additional thought actually - I wonder if we need to do different numbers for a and b. I'm curious whether other libraries try to optimize on this scenario - much like Rust Decimal does (e.g. same scale, 32 bit divisor)!

PR #416, opened 2021-08-20
Add benchmarks to compare different structures

c410-f3r:
Only compares floating primitives with arithmetic operations using equal and different left and right hand values.

add
div

Quite interesting results


paupino:
Fantastic - thank you!

PR #421, opened 2021-08-28
Initial implementation for significant figure rounding

### LINKED ISSUE #389: support rounding to significant figures instead of dp
as title. i just can't ctrl+F the word "significant" in the doc, probably some operation on exponent in decimal struct would work.

paupino:
Resolves #389


schungx:
Another `round_sf_with_strategy` might be useful, as for many calculations maybe we don't want to always round to even numbers?

paupino:
Yep, certainly possible to create a function for this behavior also - it shouldn't be too much of a lift.

PR #441, opened 2021-10-20
Feature: Retain excess precision during float conversions

### LINKED ISSUE #438: Feature request: Conversion from float retaining excess precision
Currently this code prints `0.1`:

```
let d = Decimal::from_f64(0.1).unwrap();
println!("{}", d);
```

At first that sounds like correct behavior and what one would want. But I would argue that it is in fact not the correct result, and that it should instead output `0.10000000000000000555111512313`.

As we know, the token `0.1` is read by Rust into an `f64` number, where it is represented by a floating-point number equal to the fraction `3602879701896397/36028797018963968`. (This ratio can be obtained in Python using `0.1 .as_integer_ratio()`, [or in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6d3496325cbba0808235588d3996cbfc) by printing `BigRational::from_float(0.1)`.) While the resulting number can be displayed as 0.1 without loss of precision _when again re-read as float_, and while humans often think of it as 0.1, it is the fraction that is in fact used in floating-point calculations.

This is why I would expect `"0.1".parse::<Decimal>().unwrap()` to differ from `Decimal::from_f64(0.1).unwrap()`. The former reads from decimal and is precisely `0.1`, whereas the latter converts the result of the lossy conversion of the `0.1` token to floating point. The input of `Decimal::from_f64(0.1)` is the number corresponding to the above-mentioned fraction, which has an exact decimal representation (because its denominator is a power of 2): `0.1000000000000000055511151231257827021181583404541015625`. Given the constraint of a mantissa smaller than `2**96`, the closest `rust_decimal::Decimal` representation of this is `10000000000000000555111512313 * 10**-29`, so I'd expect the `Decimal` to display as `0.10000000000000000555111512313`.

The same applies to `from_f32` where `Decimal` has enough precision to hold all the digits of [13421773/134217728](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6ee605efebfccdeb2db2edd3012ab848), and should display `0.100000001490116119384765625`.

Python's `decimal` does the equivalent conversion, except its decimal is not limited to u128, so it can accommodate the whole float:

```
>>> import decimal
>>> decimal.Decimal(0.1)
Decimal('0.1000000000000000055511151231257827021181583404541015625')
>>> decimal.Decimal("0.1")
Decimal('0.1')
```

Looking at the implementation, it's not at all obvious that the current behavior is intended. The intention seems to be to convert the number with maximum possible precision. While I'm aware that some people might consider the current behavior to be a feature, I'd still ask to change it, or at least to provide a different method for converting floats to decimals while minimizing the difference from the original number.

paupino:
Fixes #438 

Additional functions `from_f32_retain` and `from_f64_retain` created to retain additional float bits while parsing `f32`/`f64` respectively.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1418,9 +1418,9 @@ impl Decimal {
     /// assert_eq!("0.1", Decimal::from_f32(0.1_f32).unwrap().to_string());
     ///
     /// // Sometimes, we may want to represent the approximation exactly.
-    /// assert_eq!("0.100000001490116119384765625", Decimal::from_f32_retain_excess_bits(0.1_f32).unwrap().to_string());
+    /// assert_eq!("0.100000001490116119384765625", Decimal::from_f32_retain(0.1_f32).unwrap().to_string());
     /// ```
-    pub fn from_f32_retain_excess_bits(n: f32) -> Option<Self> {
+    pub fn from_f32_retain(n: f32) -> Option<Self> {
         from_f32(n, false)
     }
 
@@ -1439,9 +1439,9 @@ impl Decimal {
     /// assert_eq!("0.1", Decimal::from_f64(0.1_f64).unwrap().to_string());
     ///
     /// // Sometimes, we may want to represent the approximation exactly.
-    /// assert_eq!("0.1000000000000000055511151231", Decimal::from_f64_retain_excess_bits(0.1_f64).unwrap().to_string());
+    /// assert_eq!("0.1000000000000000055511151231", Decimal::from_f64_retain(0.1_f64).unwrap().to_string());
     /// ```
-    pub fn from_f64_retain_excess_bits(n: f64) -> Option<Self> {
+    pub fn from_f64_retain(n: f64) -> Option<Self> {
         from_f64(n, false)
     }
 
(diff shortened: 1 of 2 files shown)


hniksic:
Thanks for implementing this. As mentioned in #438, please do consider renaming the functions to `from_f32_retain` and `from_f64_retain`, which are nicer/shorter but still descriptive.

I tried this branch with several numbers, included the 0.1 example, and the results are as expected - `Decimal::from_f32_retain_excess_bits(0.1)` returns `0.100000001490116119384765625` and `Decimal::from_f64_retain_excess_bits(0.1)` returns `0.1000000000000000055511151231`.

The `f64` result looked at first like it lacked a digit - it returns a decimal with `mantissa()` and `scale()` of `1000000000000000055511151231` and `28` respectively, where I expected `10000000000000000555111512313` and `29`. But then I learned that the scale is capped at 28, so this is actually correct.

Out of curiosity, why is the maximum scale 28? The number of available bits after taking account the mantissa and sign is quite large (31 bits, allowing scales up to a billion). I understand that the limit corresponds to the maximum scale of the mantissa, but it seems to unnecessarily limit the significant precision of numbers closer to zero.

PR #445, opened 2021-11-17
Add support for Rocket forms

Misterio77:
Hello!

Here's a simple implementation for [Rocket](https://rocket.rs)'s [`FromFormField`](https://api.rocket.rs/v0.5-rc/rocket/form/trait.FromFormField.html).

This allows reading `Decimal`s straight from web form fields (either bodies or query strings).

I'm not sure if the feature name is good enough, but i couldn't think of any other. I've also added a test. Should i add a failing one as well?


src/rocket.rs
```diff
@@ -0,0 +1,12 @@
+use rocket::form::{self, FromFormField, ValueField};
+use std::str::FromStr;
+use crate::Decimal;
+
+impl<'v> FromFormField<'v> for Decimal {
+    fn default() -> Option<Self> {
+        Default::default()
+    }
+    fn from_value(field: ValueField<'v>) -> form::Result<'v, Self> {
+        Decimal::from_str(field.value).map_err(|_| form::Error::validation("not a valid number").into())
```
paupino, line 10: Newbie question: do you need to do anything to handle empty strings? Or is that handled by Rocket if you mark the `Decimal` as `Option`?

Cargo.toml
```diff
@@ -51,6 +52,7 @@ default = ["serde", "std"]
 legacy-ops = []
 maths = []
 maths-nopanic = ["maths"]
+rocket-fromform = ["rocket"]
```
paupino, line 55: Maybe just `rocket-form`? It'd be nice to be able to extend it for other form features if they ever became available.

paupino (commented):
Sorry for the delay in review! I like this - in fact, it's something I've been meaning to look into since I've been playing with Rocket for some personal projects.

My main comment is perhaps simplifying the feature name. It's really a nit, however I think it'll broaden the scope for any future functionality.

src/rocket.rs
Misterio77: The latter.

If you require a `Decimal` and `from_value` fails (because of an empty or invalid value), the request fails (similary to how other types can fail, such as ints). If you require an `Option<Decimal>`, values that fail `from_value` become a `None`.

Cargo.toml
Misterio77: What about `rocket-traits`? That should let us implement other stuff, such as [`FromParam`](https://api.rocket.rs/v0.5-rc/rocket/request/trait.FromParam.html).

paupino: Even better. I like that.

src/rocket.rs
Misterio77: Oops my mistake. I've implemented `default`, so when `from_value` (if running on lenient mode, which is the default one) fails it is called and returns `Decimal::default()`, which was not intended and could be misleading.

I'm gonna go ahead and remove that

paupino (approved):
The _only_ thing missing is running the tests within CI. This can be done by modifying `Makefile.toml` - perhaps by creating a new test task and then adding it to `test-all`.  (Update: You can probably copy `test-legacy-ops` as an example)

This just ensure that the tests are run on commit and future changes won't accidentally break this feature.

paupino:
This should help with testing the CI checks too: https://github.com/paupino/rust-decimal/blob/master/BUILD.md#formatting--code-style

PR #454, opened 2021-12-29
Attempt to fix `from_str` overflow rounding

### LINKED ISSUE #453: `from_str` rounding issues when overflow from too many digits

```rust
assert_eq!(
);
```

chris-cantor:
resolves #453

The culprit is the rounding behaviour prevented it from overflowing, and ended up returning an `Ok` in the end. The fix is: we should not attempt to round a number before the decimal point.

Having read through the code, I think there are opportunities to make `parse_str_radix_10` faster & simpler, e.g. removing the coeff `ArrayVec` and doing it in one pass; using 128 bit integer arithmetic etc.

I wonder what is your view on this?

### DIFF
--- a/src/ops/array.rs
+++ b/src/ops/array.rs
@@ -92,6 +92,24 @@ pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
     carry as u32
 }
 
+pub(crate) fn add_by_internal_flattened(value: &mut [u32; 3], by: u32) -> u32 {    let mut carry: u64;
+    let mut sum: u64;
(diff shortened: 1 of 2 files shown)


chris-cantor:
I tried to remove coeff `ArrayVec` and perform calculation in a single pass.
Fixed the issues and coincidentally performance seemed to improve quite a bit!

```
cargo bench --bench lib_benches -- decimal_from_str

master:

7289df8:
test decimal_from_str        ... bench:         193 ns/iter (+/- 1)
```

chris-cantor:
Closed in favour of #455

PR #455, opened 2021-12-30
Optimize `parse_str_radix_10`

### LINKED ISSUE #453: `from_str` rounding issues when overflow from too many digits

chris-cantor:
Replacing #454, this PR resolves #453 and:

1. Added some tests
2. Fixed two edge cases
3. Simplified the logic of `parse_str_radix_10`
4. Optimized the performance with 64/128 bit integers
5. Speed up further with tail call optimization

Running `cargo bench --bench lib_benches -- decimal_from_str`

On M1 Pro, the numbers are:

```
master:

7f18e0d:
test decimal_from_str        ... bench:         88  ns/iter (+/- 21)
```

On desktop Ryzen, the numbers are:
```
master:
test decimal_from_str        ... bench:         557 ns/iter (+/- 13)

7f18e0d:
test decimal_from_str        ... bench:         125 ns/iter (+/- 6)
```


paupino:
This looks impressive - upping the minimum compiler version to leverage const generics is already paying off by the looks! 

Since the change is quite large I want to spend a bit of time reviewing it over the next day or two. I want to get a better understanding of the `#[cold]` attributes in particular and make sure they make sense in this context.

In lieu of review, thank you for such a thoroughly put together PR!

chris-cantor:
Thank you for the quick response! No problem, definitely take your time. Just ping us if parts of the code seems unclear to you.

PR #456, opened 2021-12-31
Optimize `parse_str_radix_10`

### LINKED ISSUE #453: `from_str` rounding issues when overflow from too many digits

chris-cantor:
Replacing #454, this PR resolves #453 and:

1. Added some tests
2. Fixed two edge cases
3. Simplified the logic of `parse_str_radix_10`
4. Optimized the performance with 64/128 bit integers
5. Speed up further with tail call optimization

Running `cargo bench --bench lib_benches -- decimal_from_str`

On M1 Pro, the numbers are:
```
master:

7f18e0d:
test decimal_from_str        ... bench:         88  ns/iter (+/- 21)
```

On Ryzen 3990x, the numbers are:
```
master:
test decimal_from_str        ... bench:         557 ns/iter (+/- 13)

432f9d6:
decimal_from_str             ... bench:         129 ns/iter (+/- 2)
```

I know this is a lot of code changes. No rush at all and feel free to ask us questions if pieces of the code seems unclear.

Have a happy new year too!

### DIFF
--- a/src/ops/array.rs
+++ b/src/ops/array.rs
@@ -54,6 +54,7 @@ pub(crate) fn rescale_internal(value: &mut [u32; 3], value_scale: &mut u32, new_
     }
 }
 
+#[cfg(feature = "legacy-ops")]
 pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
     let mut carry: u64 = 0;
     let vl = value.len();
@@ -88,6 +89,25 @@ pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
         }
     } else {
         panic!("Internal error: add using incompatible length arrays. {} <- {}", vl, bl);
+    }
+    carry as u32
+}
+
+pub(crate) fn add_by_internal_flattened(value: &mut [u32; 3], by: u32) -> u32 {
+    let mut carry: u64;
+    let mut sum: u64;
+    sum = u64::from(value[0]) + u64::from(by);
+    value[0] = (sum & U32_MASK) as u32;
+    carry = sum >> 32;
+    if carry > 0 {
+        sum = u64::from(value[1]) + carry;
+        value[1] = (sum & U32_MASK) as u32;
+        carry = sum >> 32;
+        if carry > 0 {
+            sum = u64::from(value[2]) + carry;
+            value[2] = (sum & U32_MASK) as u32;
+            carry = sum >> 32;
+        }
     }
     carry as u32
 }
--- a/src/str.rs
+++ b/src/str.rs
@@ -632,86 +632,105 @@ mod test {
 
     #[test]
     fn from_str_rounding_0() {
-        assert_eq!(parse_str_radix_10("1.234").unwrap(), Decimal::new(1234, 3));
+        assert_eq!(
+            parse_str_radix_10("1.234").unwrap().unpack(),
+            Decimal::new(1234, 3).unpack()
+        );
     }
 
     #[test]
     fn from_str_rounding_1() {
         assert_eq!(
-            parse_str_radix_10("11111_11111_11111.11111_11111_11111").unwrap(),
-            Decimal::from_i128_with_scale(11_111_111_111_111_111_111_111_111_111, 14)
+            parse_str_radix_10("11111_11111_11111.11111_11111_11111")
+                .unwrap()
+                .unpack(),
+            Decimal::from_i128_with_scale(11_111_111_111_111_111_111_111_111_111, 14).unpack()
         );
     }
 
     #[test]
     fn from_str_rounding_2() {
         assert_eq!(
-            parse_str_radix_10("11111_11111_11111.11111_11111_11115").unwrap(),
-            Decimal::from_i128_with_scale(11_111_111_111_111_111_111_111_111_112, 14)
+            parse_str_radix_10("11111_11111_11111.11111_11111_11115")
+                .unwrap()
+                .unpack(),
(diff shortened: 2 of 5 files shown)


TradingTomatillo:
Pasting the interesting part of a comment here so I can respond in the new PR:

```
Since the change is quite large I want to spend a bit of time reviewing it over the next day or two. I want to get a better understanding of the #[cold] attributes in particular and make sure they make sense in this context.
```

Cold basically informs LLVM that we do not expect this function will get called. Practically speaking, that means LLVM *might*

- Optimize for size instead of speed in the cold function (in practice, these two aren't that far apart anyways)
- Use a calling convention mean to push costs to the callee instead of the caller (in the parser, the cold functions are just passed a few integers anyways)
- Optimize branch layouts with the assumption cold codepaths won't get called.

Because this is all variations of *might*, the impact varies. It has much less of an impact in the common case with small strings because we statically remove the various overflow checks, so it will impact larger strings more. Pre-size-check (the BIG variable), the cold annotations took the time from ~150 ns/iter to ~140, but after it is less impactful. It doesn't look like large decimals are very well represented in the benchmarks, although it's probably reasonable to optimize for strings with less than 18 digits in practice anyways.

TradingTomatillo:
https://godbolt.org/z/6xnxsjoze you can play around with this godbolt link and see: https://godbolt.org/z/rnhb5anhP. You can see that test3 is the non-cold one so the compiler optimizes for it (forward jumps are by default predicted false on intel so the compiler makes test3 the fallthrough case)

TradingTomatillo:
For what it's worth, the tail call approach won't be great on architectures that don't support tails or want to pass parameters on the stack. I think the only relevant architecture here for rust is webassembly (and *maybe* arm32). The options I see are:
1. Just eat the recursion cost (probably not great?)
2. Retain an improved-but-tailless version for architectures outside of x86, arm64, risc-v. This would be a higher code maintenance burden since you have 3 nontrivial cases (radix-10-x64, radix-10-general, radix-any).
3. Reuse the general radix parser for other architectures. Not that fast, but there's no reason some of the easier optimizations can't be applied there at a later date

TradingTomatillo:
Latest numbers are down to 70ns/iter on M1, will benchmark on an x86 processor in the next day or two.

src/str.rs
```diff
@@ -122,160 +122,236 @@ pub(crate) fn fmt_scientific_notation(
 }
 
 // dedicated implementation for the most common case.
+#[inline]
 pub(crate) fn parse_str_radix_10(str: &str) -> Result<Decimal, crate::Error> {
-    if str.is_empty() {
-        return Err(Error::from("Invalid decimal: empty"));
+    let bytes = str.as_bytes();
+    // handle the sign
```
paupino, line 128: nit: looks like this is leftover

src/str.rs
```diff
@@ -122,160 +122,236 @@ pub(crate) fn fmt_scientific_notation(
...
-    let mut maybe_round = false;
-    while len > 0 {
-        let b = bytes[offset];
-        match b {
-            b'0'..=b'9' => {
-                coeff.push(u32::from(b - b'0'));
-                offset += 1;
-                len -= 1;
+#[inline]
+fn byte_dispatch_u64<const POINT: bool, const NEG: bool, const HAS: bool, const BIG: bool, const FIRST: bool>(
```
paupino, line 184: My one qualm with these const generics is perhaps knowing what each does. Even though this is an internal method it'd be great to add a little bit of documentation just to cover what the purpose of each is for. 
* `POINT` - a decimal point has been seen
* `NEG` - we've encountered a `-` and the number is negative
* `HAS` - a digit has been encountered (when `HAS` is `false` it's invalid)
* `BIG` - a number that uses 96 bits instead of only 64 bits
* `FIRST` - true if it is the first byte in the string

src/ops/array.rs
```diff
@@ -92,6 +93,25 @@ pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
     carry as u32
 }
 
+pub(crate) fn add_by_internal_flattened(value: &mut [u32; 3], by: u32) -> u32 {
```
paupino, line 96: I previously had something similar - `add_by_internal3`. At some stage I'd love to move away from these functions however it makes sense to introduce it for this PR.

paupino (commented):
This is quite an innovative approach to this!

I see how the tail recursion could be problematic if optimization isn't available/used. Would it be worthwhile to feature flag this in the meantime? That way it could be an opt in (or opt out) feature. 

You're right that it does have the overhead of maintaining two different versions however we should be able to _slightly_ cater for that with CI. It's not without other downsides of course (e.g. discovery, deprecation) but for now it may make sense.

Thoughts? I can always help wrap up the PR with feature flags + CI if you like.

chris-cantor:
Sure! Let me work on the changes

TradingTomatillo:
As for the tail-call-or-not question: Everything before the tail calls is still a significant improvement and shouldn't really introduce a lot of platform dependency.

For the tail calls, it might make sense to keep the later commits here as a WIP-PR, add the non-tail optimizations to the generalized parser, and see what performance looks like there. If that's performing acceptably choice 3 (only use good one on x86/arm/riscv) might be fine?

paupino:
That makes sense. I'm going to merge this as is for now, and we can iterate from there.

paupino:
Ok, that's weird. This is merged now in fca96a533e675e9953addcaeaa836d556786ff5e? I'm going to close this PR since it remained open.

PR #460, opened 2022-01-09
Maintenance

c410-f3r:


src/mysql.rs
```diff
@@ -26,7 +26,7 @@ mod diesel_mysql {
             // internal types.
             let bytes = numeric.ok_or("Invalid decimal")?;
             let s = std::str::from_utf8(bytes)?;
-            Decimal::from_str(&s).map_err(|e| e.into())
+            Decimal::from_str(s).map_err(|e| e.into())
```
c410-f3r, line 29: Clippy stuff

src/postgres.rs
```diff
@@ -460,7 +460,7 @@ mod diesel_postgres {
 }
 
 #[cfg(any(feature = "db-postgres", feature = "db-tokio-postgres"))]
-mod postgres {
+mod _postgres {
```
c410-f3r, line 463: Clippy stuff. Feel free to choose another name if desired

src/serde.rs
```diff
@@ -94,7 +94,7 @@ pub mod float {
 ///
 /// #[derive(Serialize, Deserialize)]
 /// pub struct StringExample {
-///     #[serde(with = "rust_decimal::serde::string")]
+///     #[serde(with = "rust_decimal::serde::str")]
```
c410-f3r, line 97: CI is not currently testing the `serde-with-str` feature

src/serde.rs
```diff
@@ -120,7 +124,9 @@ pub mod str {
     where
         S: serde::Serializer,
     {
-        value.to_string().serialize(serializer)
+        ArrayString::<MAX_STR_BUFFER_SIZE>::try_from(format_args!("{}", value))
+            .map_err(S::Error::custom)?
+            .serialize(serializer)
```
c410-f3r, line 129: There is no need to perform a heap allocation.

`to_string` comes from `Display`, which is the same of `format_args!("{}", value)`.

src/ops/array.rs
```diff
@@ -94,32 +98,35 @@ pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
...
-        sum = (*i as u64) + carry;
-        *i = (sum & U32_MASK) as u32;
-        carry = sum >> 32;
+pub(crate) fn add_one_internal(value: &mut [u32; 3]) -> u32 {
+    manage_add_by_internal(1, value)
+}
+
+// `u64 as u32` are safe because of widening and 32bits shifts
+#[inline]
+pub(crate) fn manage_add_by_internal<const N: usize>(initial_carry: u32, value: &mut [u32; N]) -> u32 {
```
c410-f3r, line 111: Used by `add_one_internal` and `add_by_internal_flattened` as both are almost the same. Other similar functions could also leverage this common entry point to remove duplication.

src/ops/array.rs
```diff
@@ -94,32 +98,35 @@ pub(crate) fn add_by_internal(value: &mut [u32], by: &[u32]) -> u32 {
...
+
+    if let Some(idx) = iter.next() {
+        sum_fn(&mut carry, idx);
+    }
+
+    for idx in iter {
+        if carry > 0 {
+            sum_fn(&mut carry, idx);
+        }
     }
```
c410-f3r, line 130: https://godbolt.org/z/qK5qzEK81

Both old and new versions have the same assembly output (taking aside an initial variable), i.e., the compiler is unrolling the loop. IMO, this new version is much more readable and flexible for any `N`.

cc @chris-cantor @TradingTomatillo

src/constants.rs
```diff
@@ -9,7 +9,7 @@ pub const UNSIGN_MASK: u32 = 0x4FFF_FFFF;
 // contain a value between 0 and 28 inclusive.
 pub const SCALE_MASK: u32 = 0x00FF_0000;
 pub const U8_MASK: u32 = 0x0000_00FF;
-pub const U32_MASK: u64 = 0xFFFF_FFFF;
+pub const U32_MASK: u64 = u32::MAX as _;
```
c410-f3r, line 12: More readable IMO

src/ops/array.rs
```diff
@@ -15,42 +15,46 @@ pub(crate) fn rescale_internal(value: &mut [u32; 3], value_scale: &mut u32, new_
...
     if *value_scale > new_scale {
-        let mut diff = *value_scale - new_scale;
+        let mut diff = value_scale.wrapping_sub(new_scale);
         // Scaling further isn't possible since we got an overflow
         // In this case we need to reduce the accuracy of the "side to keep"
 
         // Now do the necessary rounding
         let mut remainder = 0;
-        while diff > 0 {
+        while let Some(diff_minus_one) = diff.checked_sub(1) {
```
c410-f3r, line 24: The same branch condition of `diff > 0`. Removes one arithmetic operation and assembly output shouldn't change

src/ops/array.rs
```diff
@@ -15,42 +15,46 @@ pub(crate) fn rescale_internal(value: &mut [u32; 3], value_scale: &mut u32, new_
     }
 
     if *value_scale > new_scale {
-        let mut diff = *value_scale - new_scale;
+        let mut diff = value_scale.wrapping_sub(new_scale);
```
c410-f3r, line 18: `wrapping_sub` is explicitly telling the intended arithmetic operation behavior, i.e., it is self documented. This will help external users to have a better glimpse of what is going on

src/ops/array.rs
```diff
@@ -209,7 +216,7 @@ pub(crate) fn mul_part(left: u32, right: u32, high: u32) -> (u32, u32) {
 }
 
 // Returns remainder
-pub(crate) fn div_by_u32(bits: &mut [u32], divisor: u32) -> u32 {
+pub(crate) fn div_by_u32<const N: usize>(bits: &mut [u32; N], divisor: u32) -> u32 {
     if divisor == 0 {
         // Divide by zero
         panic!("Internal error: divide by zero");
```
c410-f3r, line 222: Should probably sign an overflow instead of aborting the thread

src/ops/array.rs
```diff
@@ -271,7 +278,7 @@ pub(crate) fn cmp_internal(left: &[u32; 3], right: &[u32; 3]) -> core::cmp::Orde
 }
 
 #[inline]
-pub(crate) fn is_all_zero(bits: &[u32]) -> bool {
+pub(crate) fn is_all_zero<const N: usize>(bits: &[u32; N]) -> bool {
     bits.iter().all(|b| *b == 0)
 }
```
c410-f3r, line 283: I **think** this will help the compiler unroll the loop. Not sure if it is already optimizing but such addition doesn't hurt.

The same reasoning applies to the other `<const N: usize>` functions

src/ops/array.rs
chris-cantor: Yes, I think after we optimized `from_str_radix_n` these functions could be removed.

src/ops/array.rs
chris-cantor: Since this function is inline, it actually was unrolled (I checked on master).

src/serde.rs
paupino: Oh, interesting. This is the documentation tests in particular... I wonder if this would've been picked up in publish?

I may see if we can capture this. Less blind spots the better!

src/postgres.rs
paupino: I had this in another project too. for now this should be fine since it's a private mod.

paupino:
Thank you for doing this. Reducing CI warnings is a good thing and will make things easier going forward!

PR #464, opened 2022-01-16
Add saturated variants

c410-f3r:
Not sure if you want these methods but it is reasonable to assume that they are important. As for myself, this PR will prevent me from `unwrap_or` everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,11 +229,11 @@ Please refer to the [Build document](BUILD.md) for more information on building 
 
 ## Minimum Rust Compiler Version
 
-The current _minimum_ compiler version is [`1.51.0`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1510-2021-03-25)
-which was released on `2021-03-25` and included support for "const generics".
+The current _minimum_ compiler version is [`1.54.0`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1540-2021-07-29)
+which was released on `2021-07-29`.
 
 ### Updating the minimum supported version
 
-This library maintains support for rust compiler versions that are 5 minor versions away from the current stable rust compiler version.
-For example, if the current stable compiler version is `1.50.0` then we will guarantee support up to and including `1.45.0`.
+This library maintains support for rust compiler versions that are 4 minor versions away from the current stable rust compiler version.
+For example, if the current stable compiler version is `1.50.0` then we will guarantee support up to and including `1.46.0`.
 Of note, we will only update the minimum supported version if and when required.
--- a/src/arithmetic_impls.rs
+++ b/src/arithmetic_impls.rs
@@ -1,0 +1,312 @@
+// #[rustfmt::skip] is being used because `rustfmt` poorly formats `#[doc = concat!(..)]`. See
+// https://github.com/rust-lang/rustfmt/issues/5062 for more information.
+
+use crate::{decimal::CalculationResult, ops, Decimal};
+use core::ops::{Add, Div, Mul, Rem, Sub};
+use num_traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedRem, CheckedSub};
+
+// Macros and `Decimal` implementations
+
+#[rustfmt::skip]
+macro_rules! impl_checked {
+    ($long:literal, $short:literal, $fun:ident, $impl:ident) => {
+        #[doc = concat!(
+            "Checked ",
+            $long,
+            ". Computes `self ",
+            $short,
+            " other`, returning `None` if overflow occurred."
+        )]
+        #[inline(always)]
+        #[must_use]
+        pub fn $fun(self, other: Decimal) -> Option<Decimal> {
+            match ops::$impl(&self, &other) {
+                CalculationResult::Ok(result) => Some(result),
+                _ => None,
+            }
+        }
+    };
+}
+
+macro_rules! impl_checked_and_saturating {
+    (
+        $checked_long:literal,
+        $checked_short:literal,
+        $checked_fun:ident,
+        $checked_impl:ident,
+
+        $saturating_fun:ident,
+        $saturating_cmp:ident
+    ) => {
+        impl_checked!($checked_long, $checked_short, $checked_fun, $checked_impl);
+        impl_saturating!($saturating_fun, $checked_fun, $saturating_cmp);
+    };
+}
+
+#[rustfmt::skip]
+macro_rules! impl_saturating {
+    ($fun:ident, $impl:ident, $cmp:ident) => {
+        #[doc = concat!(
+            "Like [Decimal::",
+            stringify!($impl),
+            "] but instead of `None`, returns lower or upper bounds",
+        )]
+        #[inline(always)]
+        #[must_use]
+        pub fn $fun(self, other: Decimal) -> Decimal {
+            if let Some(elem) = self.$impl(other) {
+                elem
+            } else {
+                $cmp(&self, &other)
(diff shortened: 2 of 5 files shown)


src/arithmetic_impls.rs
```diff
@@ -0,0 +1,294 @@
+use crate::{decimal::CalculationResult, ops, Decimal};
```
c410-f3r, line 1: `src/decimals.rs` has more than two thousand line so in order to keep my mental health, I moved all arithmetic implementations to this module.

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,294 @@
...
+        /// ```
+        checked_add,
+        add_impl,
+        /// ```rust
+        /// use rust_decimal::Decimal;
+        ///
+        /// assert_eq!(Decimal::MAX.saturating_add(Decimal::ONE), Decimal::MAX);
+        /// assert_eq!(Decimal::MIN.saturating_add(Decimal::MIN), Decimal::MIN);
+        /// ```
+        saturating_add
```
c410-f3r, line 75: The `saturating` prefix is the prefix used by the standard library, i.e., probably the canonical naming

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,294 @@
...
+impl Decimal {
+    impl_checked_and_saturating!(
+        /// Checked addition. Computes `self + other`, returning `None` if overflow occurred.
+        ///
+        /// ```rust
+        /// use rust_decimal::Decimal;
+        ///
+        /// assert_eq!(Decimal::MAX.checked_add(Decimal::MAX), None);
+        /// assert_eq!(Decimal::MIN.checked_add(Decimal::MIN), None);
+        /// ```
```
c410-f3r, line 66: These macros can be de-duplicated in the near future using Rust 1.54 -> https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,294 @@
...
+        ///
+        /// ```rust
+        /// use rust_decimal::Decimal;
+        ///
+        /// assert_eq!(Decimal::MAX.checked_div(Decimal::ZERO), None);
+        /// assert_eq!(Decimal::MIN.checked_div(Decimal::ZERO), None);
+        /// ```
+        checked_div,
+        div_impl
+    );
```
c410-f3r, line 130: At lest for now, `saturating_div` is being skipped

src/arithmetic_impls.rs
paupino: I'm almost wondering if we upgrade to a minimum rust version of `1.54` preemptively? You're right - it's just around the corner, though I do think it'd help make these macros a little more readable. Admittedly, on first glance I thought the saturating alternative was missing documentation when it is actually present in the macro (or should be).

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,294 @@
+use crate::{decimal::CalculationResult, ops, Decimal};
+use core::ops::{Add, Div, Mul, Rem, Sub};
+use num_traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedRem, CheckedSub};
+
+// Macros and `Decimal` implementations
+
+macro_rules! impl_checked {
+    ($(#[$($doc:meta)*])* $fun:ident, $impl:ident) => {
+        /// Like the checked version but returns lower or upper bounds instead of `None`.
```
paupino, line 9: I think this should be under `impl_saturated` right?

paupino (commented):
I think the saturating doc is in the wrong place, but more holistically, I wonder if we should bump the min version to leverage the doc macro features?

src/arithmetic_impls.rs
c410-f3r: Well, it is up to you but an early 1.54 would be very handy.

FWICT, the Rust team does not back-port security bugs to old stable versions. For example, from https://www.cvedetails.com/vulnerability-list/vendor_id-19029/product_id-48677/Rust-lang-Rust.html it is possible to see that https://www.cvedetails.com/cve/CVE-2018-1000810/ is present from 1.26 to 1.29 but only 1.29.1 onwards has a fixing patch.

src/arithmetic_impls.rs
c410-f3r: Ops, my bad

c410-f3r:
The code has refactored to use the very useful `#[doc = concat!(..)]` feature along side more unit tests. If 1.59 is a must, then this PR can linger a little more or a new branch can be created.

It is also worth noting that `#[doc = concat!(..)]` can be used in other places to enhance documentation.

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,255 @@
...
+        assert_eq!(Decimal::MAX.saturating_sub(Decimal::MIN), Decimal::MAX);
+
+        assert_eq!(Decimal::MIN.checked_add(Decimal::MIN), None);
+        assert_eq!(Decimal::MIN.checked_div(Decimal::ZERO), None);
+        assert_eq!(Decimal::MIN.checked_mul(Decimal::MIN), None);
+        assert_eq!(Decimal::MIN.checked_rem(Decimal::ZERO), None);
+        assert_eq!(Decimal::MIN.checked_sub(Decimal::MAX), None);
+
+        assert_eq!(Decimal::MIN.saturating_add(Decimal::MIN), Decimal::MIN);
+        assert_eq!(Decimal::MIN.saturating_mul(Decimal::MIN), Decimal::MIN);
```
paupino, line 252: As above - I think this should actually result in `MAX` since two negatives multiplied would be a large number. Multiplication should perhaps have some sign combinations just to make sure it's handled correctly.

paupino (changes requested):
Personally, the version bump is worth it for the clarity in the macro generation. There are a couple of sign tests that I think are worthwhile implementing - at the very least to assert what is correct behavior.

src/arithmetic_impls.rs
```diff
@@ -0,0 +1,255 @@
...
+            stringify!($impl),
+            "] but instead of `None`, returns lower or upper bounds",
+        )]
+        #[inline(always)]
+        #[must_use]
+        pub fn $fun(self, other: Decimal) -> Decimal {
+            if let Some(elem) = self.$impl(other) {
+                elem
+            } else {
+                if self.is_sign_positive() {
```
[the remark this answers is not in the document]
c410-f3r: Thank you for taking a closer look into these operations as they were incorrect.

Now everything should be fixed and even more tests were added to ensure correctness.

README.md
```diff
@@ -229,11 +229,11 @@ Please refer to the [Build document](BUILD.md) for more information on building
 
 ## Minimum Rust Compiler Version
 
-The current _minimum_ compiler version is [`1.51.0`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1510-2021-03-25)
+The current _minimum_ compiler version is [`1.54.0`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1510-2021-03-25)
 which was released on `2021-03-25` and included support for "const generics".
```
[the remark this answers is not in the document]
c410-f3r: Thanks

paupino:
Cool. I have some minor modification however I'll merge this and then tweak afterwards (e.g. ordering of docs etc).

paupino:
Thanks for all your work on this!

c410-f3r:
Thank you very much @paupino

PR #472, opened 2022-01-31
Version 1.21

paupino:
Includes a bit of a documentation clean up with further examples added. I'm opting to start using `dec!` in doc examples too, however that can be rolled out slowly. The downside is the extra import line each time - I wonder if there is a way to import for all doc tests?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,8 @@ license = "MIT"
 name = "rust_decimal"
 readme = "./README.md"
 repository = "https://github.com/paupino/rust-decimal"
-version = "1.20.0"
+version = "1.21.0"
+build = "build.rs"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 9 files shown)


src/decimal.rs
```diff
@@ -156,25 +156,125 @@ pub enum RoundingStrategy {
 #[allow(dead_code)]
 impl Decimal {
     /// The smallest value that can be represented by this decimal type.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    /// ```
+    /// use rust_decimal::Decimal;
```
paupino, line 164: These are a bit verbose. Try out excluding these (`#`).

Cargo.toml
```diff
@@ -39,10 +40,10 @@ bytes = "1.0"
 criterion = { default-features = false, version = "0.3" }
 csv = "1"
 futures = "0.3"
-serde_derive = "1.0"
+rust_decimal_macros = { path = "macros" } # This should be ok since it's just for tests
```
paupino, line 43: Confirm macro tests/docs don't break.

PR #478, opened 2022-02-17
Borsh Support

jkbpvsc:
Adding support for Borsh serialization via an optional crate feature.
I've only added this to the Decimal object so far, please let me know where else to add this needs to be supported.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
+borsh = { default-features = false, optional = true, version = "0.9.3" }
 
 [dev-dependencies]
 bincode = "1.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ all-features = true
 [dependencies]
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
+borsh = { default-features = false, optional = true, version = "0.9.3" }
 byteorder = { default-features = false, optional = true, version = "1.3" }
 bytes = { default-features = false, optional = true, version = "1.0" }
 diesel = { default-features = false, optional = true, version = "1.4" }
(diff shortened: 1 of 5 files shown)


src/decimal.rs
```diff
@@ -19,6 +19,8 @@ use diesel::sql_types::Numeric;
 #[cfg(not(feature = "std"))]
 use num_traits::float::FloatCore;
 use num_traits::{FromPrimitive, Num, One, Signed, ToPrimitive, Zero};
+#[cfg(feature = "borsh")]
+use borsh::{BorshDeserialize, BorshSerialize};
```
paupino, line 23: Small nit: you could just add the qualifier on line 102 and get rid of this.
i.e.
```
#[cfg_attr(feature = "borsh", derive(borsh::BorshDeserialize, borsh::BorshSerialize))]
```

Cargo.toml
```diff
@@ -33,6 +33,7 @@ rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
+borsh = { default-features = false, optional = true, version = "0.9.3" }
```
paupino, line 36: nit: Dependencies are in alphabetical order at the moment.

paupino (commented):
Thanks for the contribution!

Is there a way that this feature can be tested? Some information about how testing is done in this library can be [found here](https://github.com/paupino/rust-decimal/blob/master/BUILD.md#testing). To begin with it could be as simple as a serialize into / deserialize out of test to make sure there is an identity relationship. Perhaps something like this: https://github.com/paupino/rust-decimal/blob/master/tests/decimal_tests.rs#L121

The only other place that would be useful to add some information about this would be in `README.md` under the `features` section with perhaps a link back to the `borsh-rs` site. I'm more than happy to write this part if you prefer.

jkbpvsc:
Added a test, cleaned up the code with your suggestions, added a section in the read me, and updated the makefile to test this feature.

paupino:
Looks good! Thank you for the quick turnaround.

jkbpvsc:
My pleasure, likewise!

jkbpvsc:
Let me know when a new version is published so I update my dependencies.

paupino:
@jkbpvsc I'll be publishing this in version 1.22.0 in about 10 minutes! Thanks again.

Issue #492, opened 2022-03-14
Trait to convert from a Decimal


turion:
It would be really helpful if there was a trait like

```rust
trait FromDecimal<T> {
    fn from_decimal(x: Decimal) -> Option<T>
}
```

Alternatively, it would be great if `TryInto` was implemented for all the primitive types.

I'm aware of `ToPrimitive` and `FromPrimitive`, but it doesn't suit my needs because it is not polymorphic in the output type.

I can submit a PR if you are interested.

paupino:
I like the idea of implementing `TryInto`/`TryFrom` for each of the primitive types as opposed to a new trait, namely as it'd be easier to discover. I believe these implementations only exist for `f32`/`f64` at the moment, however it shouldn't be too tricky to expand this for other primitives. That being said, for implementing this it should likely follow the same lines as `impl_from!` (i.e. could require some refactoring of `f32` and `f64` to follow this too).

Very much open to a PR if you're willing to take a stab at it!

turion:
Ok, I opened https://github.com/paupino/rust-decimal/pull/493 and I'm open to your comments :)

PR #493, opened 2022-03-15
Add TryFrom impls to primitive types (#492)

turion:


src/decimal.rs
```diff
@@ -1694,8 +1694,68 @@ const fn flags(neg: bool, scale: u32) -> u32 {
...
+        ///
+        /// Can fail if the `Decimal` is out of range.
+        /// For integer types, truncate and return the integer part.
+        ///
+        impl core::convert::TryFrom<Decimal> for $TInto {
+            type Error = crate::Error;
+
+            #[inline]
+            fn try_from(t: Decimal) -> Result<Self, Error> {
+                $conversion_fn(&t).ok_or_else(|| Error::from(format!("Failed to convert to {}", $name)))
```
paupino, line 1710: nit: you could use `stringify!($TInto)` instead of $name here.

src/decimal.rs
```diff
@@ -1694,8 +1694,68 @@ const fn flags(neg: bool, scale: u32) -> u32 {
     (scale << SCALE_SHIFT) | ((neg as u32) << SIGN_SHIFT)
 }
 
+macro_rules! impl_try_from_decimal {
+    ($TInto:ty, $conversion_fn:path, $name:expr) => {
+        ///
+        /// Try to convert a `Decimal` to a primitive type.
```
paupino, line 1700: I wonder if you could leverage `doc` macros here - similar to `arithmetic_impl.rs`. e.g.:

```
#[doc = concat!(
            "Try to convert a `Decimal` into `",
            stringify!($TInto),
            "`. Returns an error if the `Decimal` is unable to be represented as a `",
            stringify!($TInto),
            "`."
        )]
```

You could probably extend it to also reference the integer logic too (e.g. passed through as `$additional_doc:expr` or something).

src/decimal.rs
paupino: Also, I know the precedent right now is to return a `String` for an `Error`, however it'd be preferable to introduce an new `enum` option within the `Error` type. Ideally, I'd like to move away from `ErrorString`.

src/decimal.rs
```diff
@@ -1694,8 +1694,68 @@ const fn flags(neg: bool, scale: u32) -> u32 {
...
+        /// Conversion to `Decimal`.
+        ///
+        /// Can always fail if the value is out of range for `Decimal`.
+        ///
+        impl core::convert::TryFrom<$TFrom> for Decimal {
+            type Error = crate::Error;
+
+            #[inline]
+            fn try_from(t: $TFrom) -> Result<Self, Error> {
+                $conversion_fn(t).ok_or_else(|| Error::from(format!("Failed to convert to Decimal")))
```
paupino, line 1743: Same comment as above for `ErrorString`.

src/decimal.rs
```diff
@@ -1694,8 +1694,68 @@ const fn flags(neg: bool, scale: u32) -> u32 {
...
+impl_try_from_decimal!(u8, Decimal::to_u8, "u8");
+impl_try_from_decimal!(u16, Decimal::to_u16, "u16");
+impl_try_from_decimal!(u32, Decimal::to_u32, "u32");
+impl_try_from_decimal!(u64, Decimal::to_u64, "u64");
+impl_try_from_decimal!(u128, Decimal::to_u128, "u128");
+
+macro_rules! impl_try_from_primitive {
+    ($TFrom:ty, $conversion_fn:path) => {
+        ///
+        /// Conversion to `Decimal`.
```
paupino, line 1734: Consider utilizing the `doc` macro here too. e.g. 

```
#[doc = concat!(
            "Try to convert a `",
            stringify!($TFrom),
            "` into a `Decimal`."
        )]
```

It's an interesting one for erroring here - it'll only likely error on `128` bit integers. This could be again passed through as additional doc if need be.

src/decimal.rs
```diff
@@ -1694,8 +1694,68 @@ const fn flags(neg: bool, scale: u32) -> u32 {
...
+
+impl_try_from_primitive!(f32, Self::from_f32);
+impl_try_from_primitive!(f64, Self::from_f64);
+
 macro_rules! impl_from {
     ($T:ty, $from_ty:path) => {
+        ///
+        /// Conversion to `Decimal`.
+        ///
+        /// Guaranteed to succeed without errors.
```
paupino, line 1757: This isn't quite right. In fact, it's something I probably need to look into some more. At the moment, it'll panic if it's unable to be represented however it'll panic with an ambiguous unwrap error.

tests/decimal_tests.rs
```diff
@@ -2864,6 +2864,13 @@ fn it_converts_from_f64_retaining_bits() {
     }
 }
 
+#[test]
+fn it_converts_to_integers() {
+    assert_eq!(i64::try_from(Decimal::ONE), Ok(1));
+    assert!(i64::try_from(Decimal::MAX).is_err());
```
paupino, line 2870: Would be great to assert the value of the error too.

paupino (commented):
A few nits in here, but namely because we'll be introducing a new interface. Getting it "mostly" right early is easier than changing it later!

src/decimal.rs
turion: Ok, I'll remove that comment and leave it to you.

src/decimal.rs
```diff
@@ -1694,8 +1694,74 @@ const fn flags(neg: bool, scale: u32) -> u32 {
     (scale << SCALE_SHIFT) | ((neg as u32) << SIGN_SHIFT)
 }
 
+macro_rules! integer_docs {
+    ( true ) => {", truncating and returning the integer part first"};
```
paupino, line 1698:
```suggestion
    ( true ) => {" by truncating and returning the integer component"};
```

paupino:
@turion You should be able to resolve the style issues by running `makers format` (this runs `cargo fmt -- --emit=files` behind the scenes).

turion:
The trouble here seems to be that `cargo fmt` isn't idempotent here! I've run `cargo fmt` already, and amended the change, but it always wants to indent once more.

PR #496, opened 2022-03-18
Fix Decimal::to_i64 for I64::MIN

arthurprs:
The overflow check didn't consider that the maximum mantissa for negative numbers can use the extra bit.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2127,14 +2127,14 @@ impl ToPrimitive for Decimal {
     fn to_i64(&self) -> Option<i64> {
         let d = self.trunc();
         // Quick overflow check
-        if d.hi != 0 || (d.mid & 0x8000_0000) > 0 {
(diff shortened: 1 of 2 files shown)


src/decimal.rs
```diff
@@ -2127,14 +2127,14 @@ impl ToPrimitive for Decimal {
     fn to_i64(&self) -> Option<i64> {
         let d = self.trunc();
         // Quick overflow check
-        if d.hi != 0 || (d.mid & 0x8000_0000) > 0 {
```
paupino, line 2130: This is close, but isn't quite right either I don't think. I think we need to check explicitly for `i64::MIN` since that is the value which is unable to be represented.
I'll write a couple more tests and contribute to this PR.

paupino:
Merged manually in [2de2a6d](https://github.com/paupino/rust-decimal/commit/2de2a6dd2f385e98c4019ebe38b5c6de5fef6cba)

PR #498, opened 2022-03-21
implement borsh schema

jnitard:
I would like to use rust_decimal with BorshSchema.
There’s nothing to do but deriving from an extra trait when the feature is active, so that should not impact existing users.


paupino:
Thanks for putting this together. Is there any way to test it? While it looks decent enough it’d be great to confirm it’s doing what it should be doing.

jnitard:
> Thanks for putting this together. Is there any way to test it? While it looks decent enough it’d be great to confirm it’s doing what it should be doing.

Well, there was of course. I am not using the schema that way so I did not come up with immediately.

paupino:
Awesome! Looks good

PR #501, opened 2022-03-26
Fixes no_std compile by avoiding to_string

### LINKED ISSUE #500: Latest version does not compile when the `std` feature is disabled
Offending commit: 168fa67932b88bf65d5fa6c84612f02f052e6ec5

The `TryFrom` impls use `to_string`, from the [`ToString` trait](https://doc.rust-lang.org/stable/std/string/trait.ToString.html) from the prelude. This trait isn't available when `std` is disabled.

paupino:
Fixes #500


paupino:
I'll look into better `no_std` checks in a separate PR. For now this should resolve the compile-time issue with `1.23.0`

PR #512, opened 2022-05-02
Update diesel to 2.0.0.rc-0

Guara92:
Hi, 

I open this as draft, maybe this compatibility could be a different feature to avoid breaking changes to downstram projects and released after `diesel` 2.0 release, in the mean time a branch could be useful to others experimenting upgrading `diesel`.


paupino:
Whoops, didn't mean to mark this as ready for review!

Thanks for creating this. I would be open to creating another feature if it's required, otherwise more than happy to keep this open for others until 2.0 is stabilized.

paupino:
This is now released as a feature flag within `1.26`. You can enable the `rc` version of `diesel` by using either:
* `db-diesel2-mysql`
* `db-diesel2-postgres`

Please note, it is using `rc-1` and will be considered an "unstable" feature until the official release of `diesel 2.0` however should be sufficient for your needs!

Consequently, closing this PR.

Issue #489, opened 2022-03-04
Adding a rand feature


happysalada:
Hi, thank you for the great library!

I'm going through generating some fake data with fake-rs for my tests. Some of my attributes are decimals.
The fake-rs number generation relies on implementing SimpleUniform from the rand crate.
I see that there is already support for fuzzers, I think ultimately the implementation would be similar.
I am happy to have a look, I was wondering though if it was something you would think worth considering ? (making sure I'm not missing something).

paupino:
That definitely sounds like a useful idea. In my opinion, if it makes the path to using Decimal easier then it’s worthwhile considering for a feature. In this case, allowing this to be consumed by a library like ‘fake-rs’ makes a lot of sense.

happysalada:
it looks like this can be closed now.
Thank you for answering so fast again!

PR #517, opened 2022-05-27
Add `rand` support

c410-f3r:
Fix https://github.com/paupino/rust-decimal/issues/489 (Don't know why the issue was closed)

Through a feature called `rand` that is disabled by default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ bytes = { default-features = false, optional = true, version = "1.0" }
 diesel = { default-features = false, optional = true, version = "1.4" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
+rand = { default-features = false, optional = true, version = "0.8" }
 rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -41,6 +42,7 @@ bytes = { default-features = false, version = "1.0" }
 criterion = { default-features = false, version = "0.3" }
 csv = "1"
 futures = { default-features = false, version = "0.3" }
+rand = { default-features = false, features = ["getrandom"], version = "0.8" }
```
c410-f3r, line 45: It is unfortunate that cargo forces the declaration of the same dependencies with different features.

For downstream libraries, the use of `rand_core` implies the addition of a possibly unused dependency in a production environment.
For development, `getrandom` is one of the methods available to generate entropy.

src/rand.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+#[test]
+fn has_random_decimal_instances() {
+    let mut rng = rand::rngs::OsRng;
+    let random: Decimal = rng.gen();
+    for _ in 0..100 {
+        let local_random: Decimal = rng.gen();
+        if random != local_random {
```
paupino, line 30: This is a bit of an odd test to me! Do we expect the first, say 50 iterations to generate the same number as `random`?
Another idea: would it be possible to generate 100 numbers and check they're all different?

c410-f3r: Nope, the test is more like "Is the implementation generating random number at all?".

> Another idea: would it be possible to generate 100 numbers and check they're all different?

This will make CI intermittent because sometimes (even rare) it is possible to have at least one repeated element.

c410-f3r: Let me know if the updated test improved the situation.

paupino: Fair enough - let's give it a go with the updated test and if it ends up being problematic we can adjust it. Thank you!

c410-f3r:
Thanks @paupino

lukesneeringer:
As chance would have it, I need...exactly this for something I'm working on (I'm trying to get decimal support added to [ta-rs](https://github.com/greyblake/ta-rs/pull/59)).

@paupino Would it be possible / reasonable to release a 1.24 with this (and other new features on HEAD in the last couple months) in it?

paupino:
@lukesneeringer Yep, we can do that! It'll be a smaller release, but sometimes that's just fine too 😄

paupino:
@lukesneeringer Should be released now as `1.24.0`. All the best!

PR #520, opened 2022-06-07
add rkyv support

dovahcrow:
This PR add optional rkyv support for Decimal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ diesel = { default-features = false, optional = true, version = "1.4" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
 rand = { default-features = false, optional = true, version = "0.8" }
-rkyv = { default-features = false, optional = true, version = "0.7" }
+rkyv = { default-features = false, features = ["size_32", "std"], optional = true, version = "0.7" }
 rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
@@ -61,7 +61,6 @@ default = ["serde", "std"]
 legacy-ops = []
 maths = []
 maths-nopanic = ["maths"]
-rkyv = ["rkyv/size_32", "rkyv/std", "std"]
 rkyv-safe = ["bytecheck", "rkyv", "rkyv/validation"]
 rocket-traits = ["rocket"]
 rust-fuzz = ["arbitrary"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,12 +25,14 @@ all-features = true
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
 borsh = { default-features = false, optional = true, version = "0.9.3" }
+bytecheck = { default-features= false, optional = true, version = "0.6" }
 byteorder = { default-features = false, optional = true, version = "1.3" }
 bytes = { default-features = false, optional = true, version = "1.0" }
 diesel = { default-features = false, optional = true, version = "1.4" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
 rand = { default-features = false, optional = true, version = "0.8" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
+rkyv = {version = "0.7", optional = true}
```
paupino, line 38: nit: can we do a couple of things here:
1. this list is alphabetical - can we put it before `rocket`? The parameters are also alphabeticalized.
2. Can we set `default-features = false` and include any required features explicitly?

paupino (commented):
Thanks for the contribution - seems simple enough! There are a few additional changes to get this ready for merge:

1. Can you add a test for this? Possibly follow the same style as `borsh`. You'd also need to add a new test into the Makefile and "turn it on" for CI here: https://github.com/paupino/rust-decimal/blob/master/Makefile.toml#L176
2. Add a section to the readme for this feature. It doesn't have to be much! https://github.com/paupino/rust-decimal/blob/master/README.md#borsh
3. Some minor tweaks as described in the `Cargo.toml` file in regards to `default-features` - it helps us keep build times low.

dovahcrow:
@paupino Thank you for the detailed suggestions! I've revised the PR accordingly. This project is so well maintained! I appreciate all the time & effort you put on this project.

dovahcrow:
rkyv also allows one to specify a safe api with a feature flag (https://github.com/rkyv/rkyv#example). Do you want to enable that by default? 

FYI safe API allows one to deserialize types from bytes with some additional checks (e.g. byte alignment), with some performance sacrifice.

Cargo.toml
```diff
@@ -68,7 +70,7 @@ serde-str = ["serde-with-str"]
 serde-with-arbitrary-precision = ["serde", "serde_json/arbitrary_precision", "serde_json/std"]
 serde-with-float = ["serde"]
 serde-with-str = ["serde"]
-std = []
+std = ["rkyv/std"]
```
paupino, line 73: This isn't quite right, but I see what you're trying to do. Perhaps for now, we can move the `rkyv/std` feature under `rkyv` also. It just means that the feature requires the `std` feature. I'll make a code suggestion for above.

Cargo.toml
```diff
@@ -59,6 +60,7 @@ default = ["serde", "std"]
 legacy-ops = []
 maths = []
 maths-nopanic = ["maths"]
+rkyv = ["rkyv/size_32"]
```
paupino, line 63:
```suggestion
rkyv = ["rkyv/size_32", "rkyv/std", "std"]
```

paupino (approved):
One minor change required with the feature flags (`std` isn't quite right).

> rkyv also allows one to use a safe api with a feature flag (https://github.com/rkyv/rkyv#example). Do you want to enable that by default?

Your call! I'm happy with how it is right now but if you think it's a useful feature then feel free to add it.

Cargo.toml
dovahcrow: rkyv can be used without std. I added rkyv/std in the std feature of rust_decimal

Cargo.toml
paupino:
```suggestion
std = []
```

Cargo.toml
dovahcrow: Oh I just say your last comment.

Cargo.toml
dovahcrow: I think `std = ["rkyv/std"]` means if the std feature of `rust-decimal` is turned on, also turn on the `std` feature of `rkyv`?

paupino: Unfortunately, it includes the dependency by default too. See this section: https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

I see that `"rkyv?/std"` is _actually_ what we want however it requires a minimum rust version of 1.60. Unfortunately, it'll have to wait until the min version requirements catch up: https://github.com/paupino/rust-decimal#minimum-rust-compiler-version

dovahcrow: I see. right "rkyv?/std" is the correct one.

dovahcrow:
@paupino Fixed the std issue and also added safe API support via the `rkyv-safe` feature flag.

PR #525, opened 2022-06-27
Implement `Product`

c410-f3r:
Curiously, `Decimal` implements `Sum` but doesn't implement `Product`.


paupino:
Looks good - thanks!

Issue #527, opened 2022-07-06
`Decimal::try_new` is `const fn`, but not usefully so in certain cases


SohumB:
`Decimal::try_new` is intended to be used in const contexts; but there's a glitch. The function declares itself as returning `Result<Decimal, Error>`, where `Error` is the general error type for the crate. However, some of those enums involve `String`s, and any attempt to call `try_new` and match on it in a const context results in:

```
error[E0277]: can't drop `std::vec::Vec<u8>` in const contexts    
...
note: the trait `Destruct` is implemented for `std::vec::Vec<u8>`, but that implementation is not `const`                                                                                                            
```

(This precise error is due to me turning on a truckload of other const-related features, but the basic thrust of the error remains.)

This isn't actually a real issue, in that `try_new` doesn't actually _construct_ any strings — the only error case it returns is `ScaleExceedsMaximumPrecision` which absolutely could be dropped in const contexts. The bugfix here would, however, require bumping semver, as at a minimum it would involve changing `try_new`'s type.

I'm willing to submit a PR for this, but there's at least two distinct ways to break semver here, both with good arguments for it, so I didn't want to do so without opening a discussion first.

c410-f3r:
Where the `Vec<u8>` came from? `try_new` does not perform any kind of heap allocation.

```rust
use rust_decimal::Decimal;

const N: rust_decimal::Result<Decimal> = {
    Decimal::try_new(i64::MAX, u32::MAX)
};

pub fn main() {
    dbg!(N.unwrap_err());
}
```

There are `String`s in the `Error` enum because that is what was rooted and spread across the code way before its introduction. Besides, any other type of constant error is more related to the current limitations of the compiler rather than the method itself.

```rust
use rust_decimal::Decimal;

const _: () = {
    // Nope. Destructor is not allowed
    match Decimal::try_new(i64::MAX, u32::MAX) {
        Ok(elem) => Ok(elem),
        Err(err) => Err(err),
    };

    // Nope. No const unwrap Result
    Decimal::try_new(i64::MAX, u32::MAX).unwrap_err();

    // Nope. No const unwrap Option
    Decimal::try_new(i64::MAX, u32::MAX).ok().unwrap();
};

pub fn main() {
}
```

SohumB:
[...]

PR #528, opened 2022-07-06
DRAFT: constify ops

SohumB:
This PR (is intended to) make the fns in `rust_decimal::ops` const. Additionally, since `impl const Ord` [etc] are not stable yet, for these const fns to be useful, we need to actually export them.  (And eventually, once `impl const Trait` is stabilised, deprecate and unexport them.)

I've just done `cmp` so far, as a template for what will need to be done. Is this approach one that would be reasonable in `rust_decimal`?

TODO: 
* [x] `ops::cmp_impl`
* [ ] `ops::add_impl`
* [ ] `ops::sub_impl`
* [ ] `ops::mul_impl`
* [ ] `ops::div_impl`
* [ ] `ops::rem_impl`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@ extern crate alloc;
 mod constants;
 mod decimal;
 mod error;
-mod ops;
+pub mod ops;
 mod str;
 
 // We purposely place this here for documentation ordering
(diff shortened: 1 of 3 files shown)


paupino:
I quite like the intention of this PR - effectively supporting some nightly only features to support further `const` functions. That being said, I'd like to do so without exposing some of the internal functions - namely because these are subject to change. We had a brief conversation about this on #527 however wanted to provide context to supplemental information for those looking at this PR.

PR #529, opened 2022-07-17
Raising an error upon failure to deserialize overflowing number

fredfortier:
I found this panic scenario with the Decimal::serde deserializer when fuzz testing.

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -833,7 +833,7 @@ mod test {
         pub struct OverflowExample {
             value: Decimal,
         }
-        // Going a fraction about the MAX number expecting it to fail but not panic.
+        // Going a fraction above the MAX number expecting it to fail, but not panic.
         let res: Result<Decimal, _> = Decimal::from_str(format!("{}.999999", Decimal::MAX).as_str());
         // Avoiding `#[should_panic]` to ensure we're returning an error.
         if res.is_ok() {
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -826,4 +826,18 @@ mod test {
         let deserialized: StringExample = serde_json::from_str(r#"{"value":null}"#).unwrap();
         assert_eq!(deserialized.value, original.value);
     }
+
+    #[test]
+    fn with_overflow() {
+        #[derive(Serialize, Deserialize)]
+        pub struct OverflowExample {
+            value: Decimal,
+        }
+        // Going a fraction about the MAX number expecting it to fail but not panic.
+        let res: Result<Decimal, _> = Decimal::from_str(format!("{}.999999", Decimal::MAX).as_str());
+        // Avoiding `#[should_panic]` to ensure we're returning an error.
+        if res.is_ok() {
+            panic!("Expected to fail deserialization with overflow");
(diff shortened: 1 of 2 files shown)


src/serde.rs
```diff
@@ -826,4 +826,18 @@ mod test {
...
     }
+
+    #[test]
+    fn with_overflow() {
+        #[derive(Serialize, Deserialize)]
+        pub struct OverflowExample {
+            value: Decimal,
+        }
+        // Going a fraction about the MAX number expecting it to fail but not panic.
+        let res: Result<Decimal, _> = Decimal::from_str(format!("{}.999999", Decimal::MAX).as_str());
```
fredfortier, line 837: Instead of catching the expected deserialization error, it panicked like this:
```
        thread 'serde::test::with_overflow' panicked at 'assertion failed: `(left == right)`
        left: `1`,
        right: `0`', src/str.rs:384:5
```

I would prefer being able to throw any string at Decimal::str rather than implementing a separate validation layer.

fredfortier:
Since I'm expecting you to say that the use of `debug_assert!` was a deliberate performance optimization, I went ahead an added a crude validation layer. Please let me know if I just reinvented an existing util, or feel free to comment on the approach. The intent is to capture any problems with the string, as efficiently as possible, before deserializing to Decimal.

```rust

// Calculated from Decimal::MAX display length
const MAX_DECIMAL_LEN: usize = 29;

lazy_static! {
    static ref DECIMAL_RE: Regex =
        Regex::new(r"[+-]?((?P<whole>\d+)?(?P<dot>[.]))?(?P<decimal>\d+)")
            .expect("Invalid decimal validation pattern");
}

impl FromStr for DecimalWrapper {
    type Err = anyhow::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let cap = DECIMAL_RE
            .captures(s)
            .ok_or(anyhow!("Invalid decimal s={}", s))?;
        // Run basic heuristics on the regex captures before attempting any parsing.
        let whole;
        let fract;
        // Case: 1234.4444
        if let Some(whole_) = cap.name("whole") {
            whole = whole_.as_str();
            fract = &cap["decimal"];
        // Case: .4444
        } else if let Some(dot) = cap.name("dot") {
            whole = "";
            fract = &cap["decimal"];
        // Case: 1234
        } else {
            whole = &cap["decimal"];
            fract = "";
        }
        ensure!(
            fract.len() <= DEFAULT_CURRENCY_DECIMAL_PRECISION as usize,
            "Too many numbers after the decimal len({}) > {}",
            fract,
            DEFAULT_CURRENCY_DECIMAL_PRECISION
        );
        // Effectively reducing the Decimal capacity, which is fine here since our internal threshold is lower.
        ensure!(
            whole.len() < MAX_DECIMAL_LEN,
            "Preventing Decimal overflow len({}) > {}",
            whole,
            MAX_DECIMAL_LEN
        );
        Ok(DecimalWrapper::new(Decimal::from_str(s)?))
    }
}
```

paupino:
Closing due to fix in #530

PR #532, opened 2022-08-04
Cleanup of database module

paupino:
Start of some refactoring of the database modules to hopefully make supporting duplicate diesel versions a bit easier. This also fixes some bugs in the `Makefile` - tests weren't being run as per expectations after some recent (?) refactoring.


paupino:
Merging this since it will help prevent accidental errors during diesel 2.0 testing.

PR #539, opened 2022-08-22
add rkyv Clone/Copy derivations

icy-ux:
Since Decimal is Clone/Copy, we might as well make its `rkyv::Archive`
representation also have these traits.


paupino:
Makes sense to me - thank you for the contribution!

PR #541, opened 2022-09-01
Fix rounding mantissa overflow

### LINKED ISSUE #540: rounding fails with "Invalid decimal: overflow from mantissa after rounding"
Fun edge case. The decimal `792.281625142643375935439503356` hits 

https://github.com/paupino/rust-decimal/blob/4b7176177cfd3938cecb59d9358667a8a2d4ab6d/src/str.rs#L309-L312

The correct rounding behavior can be achieved with

```diff,rust
--- a/src/str.rs
+++ b/src/str.rs
@@ -306,9 +306,15 @@ fn handle_full_128<const POINT: bool, const NEG: bool, const ROUND: bool>(
                     if digit >= 5 {
                         data += 1;

-                        // Make sure that the mantissa isn't now overflowing
+                        // If the mantissa is now overflowing, round to the next
+                        // next least significant digit and discard precision
                         if overflow_128(data) {
-                            return tail_error("Invalid decimal: overflow from mantissa after rounding");
+                            if scale == 0 {
+                                return tail_error("Invalid decimal: overflow from mantissa after rounding");
+                            }
+                            data += 4;
+                            data /= 10;
+                            return handle_data::<NEG, true>(data, scale - 1);
                         }
                     }
                 } else {
```

------

With a similar input (`7.92281625142643375935439503356`), we can hit this same error in `maybe_round`, where a comment suggests it should be unreachable:

https://github.com/paupino/rust-decimal/blob/4b7176177cfd3938cecb59d9358667a8a2d4ab6d/src/str.rs#L366-L373

The same patch can be applied there, but I wonder if unifying the two rounding paths would make sense. I think that full patch would be

```diff,rust
diff --git a/src/str.rs b/src/str.rs
index d8e68e4..dafd50a 100644
--- a/src/str.rs
+++ b/src/str.rs
@@ -303,18 +303,10 @@ fn handle_full_128<const POINT: bool, const NEG: bool, const ROUND: bool>(
                 }
 
                 if ROUND {
-                    if digit >= 5 {
-                        data += 1;
-
-                        // Make sure that the mantissa isn't now overflowing
-                        if overflow_128(data) {
-                            return tail_error("Invalid decimal: overflow from mantissa after rounding");
-                        }
-                    }
+                    maybe_round(data, next_byte, scale, POINT, NEG)
                 } else {
-                    return Err(Error::Underflow);
+                    Err(Error::Underflow)
                 }
-                handle_data::<NEG, true>(data, scale)
             } else {
                 data = next;
                 let scale = scale + POINT as u8;
@@ -355,7 +347,13 @@ fn handle_full_128<const POINT: bool, const NEG: bool, const ROUND: bool>(
 
 #[inline(never)]
 #[cold]
-fn maybe_round(mut data: u128, next_byte: u8, scale: u8, point: bool, negative: bool) -> Result<Decimal, crate::Error> {
+fn maybe_round(
+    mut data: u128,
+    next_byte: u8,
+    mut scale: u8,
+    point: bool,
+    negative: bool,
+) -> Result<Decimal, crate::Error> {
     let digit = match next_byte {
         b'0'..=b'9' => u32::from(next_byte - b'0'),
         b'_' => 0, // this should be an invalid string?
@@ -366,9 +364,16 @@ fn maybe_round(mut data: u128, next_byte: u8, scale: u8, point: bool, negative:
     // Round at midpoint
     if digit >= 5 {
         data += 1;
+
+        // If the mantissa is now overflowing, round to the next
+        // next least significant digit and discard precision
         if overflow_128(data) {
-            // Highly unlikely scenario which is more indicative of a bug
-            return tail_error("Invalid decimal: overflow when rounding");
+            if scale == 0 {
+                return tail_error("Invalid decimal: overflow from mantissa after rounding");
+            }
+            data += 4;
+            data /= 10;
+            scale -= 1;
         }
     }
```

I'm submitting this as a PR.

-----

If I've deconstructed the control flow correctly, to hit these errors you need a decimal with a prefix in `0xF00000000000000000000000..0x1000000000000000000000000` and a following digit `>= 5`. If the decimal point appears anywhere within  or directly after that prefix, this error is hit. The `maybe_round` version of the error is only hit when the decimal point is exactly at index 1.

CAD97:
fixes #540

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -868,10 +868,12 @@ mod test {
     #[test]
     fn from_str_mantissa_overflow_1() {
         // reminder:
-        assert_eq!(OVERFLOW_U96, 79228162514264337593543950336);
+        assert_eq!(OVERFLOW_U96, 79_228_162_514_264_337_593_543_950_336);
         assert_eq!(
-            parse_str_radix_10("7922816251426433759354395033.56").unwrap().unpack(),
-            Decimal::from_i128_with_scale(7922816251426433759354395034, 0).unpack()
+            parse_str_radix_10("79_228_162_514_264_337_593_543_950_33.56")
+                .unwrap()
+                .unpack(),
+            Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 0).unpack()
         );
         // This is a mantissa of OVERFLOW_U96 - 1 just before reaching the last digit.
         // Previously, this would return Err("overflow from mantissa after rounding")
@@ -881,18 +883,21 @@ mod test {
     #[test]
     fn from_str_mantissa_overflow_2() {
         assert_eq!(
-            parse_str_radix_10("79228162514264337593543950335.6"),
+            parse_str_radix_10("79_228_162_514_264_337_593_543_950_335.6"),
             Err(Error::from("Invalid decimal: overflow from mantissa after rounding"))
         );
-        // this case wants to round to 79228162514264337593543950340. which is
-        // too large for the mantissa, so should fail.
+        // this case wants to round to 79_228_162_514_264_337_593_543_950_340.
+        // (79_228_162_514_264_337_593_543_950_336 is OVERFLOW_U96 and too large
+        // to fit in 96 bits) which is also too large for the mantissa so fails.
     }
 
     #[test]
     fn from_str_mantissa_overflow_3() {
         assert_eq!(
-            parse_str_radix_10("7.92281625142643375935439503356").unwrap().unpack(),
-            Decimal::from_i128_with_scale(7922816251426433759354395034, 27).unpack()
+            parse_str_radix_10("7.9_228_162_514_264_337_593_543_950_335_6")
+                .unwrap()
+                .unpack(),
+            Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 27).unpack()
         );
         // this hits the other avoidable overflow case in maybe_round
     }
--- a/src/str.rs
+++ b/src/str.rs
@@ -928,9 +928,7 @@ mod test {
     fn from_str_mantissa_overflow_3() {
         // this hits the other avoidable overflow case in maybe_round
         assert_eq!(
-            parse_str_radix_10("7.92281625142643375935439503356")
-                .unwrap()
-                .unpack(),
+            parse_str_radix_10("7.92281625142643375935439503356").unwrap().unpack(),
             Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 27).unpack()
         );
     }
--- a/src/str.rs
+++ b/src/str.rs
@@ -361,7 +361,13 @@ fn handle_full_128<const POINT: bool, const NEG: bool, const ROUND: bool>(
 
 #[inline(never)]
 #[cold]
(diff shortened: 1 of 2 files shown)


paupino (commented):
Thank you for finding this bug and issuing a fix!

I think this looks good - I have a couple of nits that it'd be good to address first (for my own future sanity reasons!).

One additional thing I think should perhaps be added to this: could you add a couple more benchmarks for these edge cases? I don't think it should make much of a difference, however it'd good to be a little more certain about this. The explicit benchmark we'd want to check is `decimal_from_str`.

src/str.rs
```diff
@@ -860,6 +865,38 @@ mod test {
...
+        // instead of successfully rounding.
+    }
+
+    #[test]
+    fn from_str_mantissa_overflow_2() {
+        assert_eq!(
+            parse_str_radix_10("79228162514264337593543950335.6"),
+            Err(Error::from("Invalid decimal: overflow from mantissa after rounding"))
+        );
+        // this case wants to round to 79228162514264337593543950340. which is
```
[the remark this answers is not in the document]
CAD97: This is the interesting case. It does initially try to round to `79_228_162_514_264_337_593_543_950_336.`, but this mantissa is `>= OVERFLOW_U96`, so we hit the second iteration and try to round to `79_228_162_514_264_337_593_543_950_340.`, which is *also* `>= OVERFLOW_U96`. This makes me realize though that the rounding should iterate through the entire mantissa when we start with `scale=0`; this can round to `79_228_162_514_264_337_593_543_950_300.`.

src/str.rs
```diff
@@ -893,6 +898,43 @@ mod test {
...
+    fn from_str_mantissa_overflow_3() {
+        assert_eq!(
+            parse_str_radix_10("7.9_228_162_514_264_337_593_543_950_335_6")
+                .unwrap()
+                .unpack(),
+            Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 27).unpack()
+        );
+        // this hits the other avoidable overflow case in maybe_round
+    }
+
```
paupino, line 937:
```suggestion
    #[test]
    fn from_str_mantissa_overflow_3() {
        // this hits the other avoidable overflow case in maybe_round
        assert_eq!(
            parse_str_radix_10("7.92281625142643375935439503356")
                .unwrap()
                .unpack(),
            Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 27).unpack()
        );
    }

    #[ignore]
    #[test]
    fn from_str_mantissa_overflow_4() {
        // Same test as above, however with underscores. This causes issues.
        assert_eq!(
            parse_str_radix_10("7.9_228_162_514_264_337_593_543_950_335_6")
                .unwrap()
                .unpack(),
            Decimal::from_i128_with_scale(79_228_162_514_264_337_593_543_950_34, 27).unpack()
        );
    }

```

src/str.rs
```diff
@@ -893,6 +898,55 @@ mod test {
...
+        // to fit in 96 bits) which is also too large for the mantissa so fails.
+    }
+
+    #[test]
+    fn from_str_mantissa_overflow_3() {
+        // this hits the other avoidable overflow case in maybe_round
+        assert_eq!(
+            parse_str_radix_10("7.92281625142643375935439503356")
+                .unwrap()
+                .unpack(),
```
paupino, line 933:
```suggestion
            parse_str_radix_10("7.92281625142643375935439503356").unwrap().unpack(),
```

Issue #49, opened 2017-10-05
Review error handling within Decimal


paupino:
There are some areas where `panic` is warranted however there are others (e.g. `rescale`, `from_bytes_le`) that we could utilize Rust's error handling further. In addition, with libraries such as `error-chain` we can perhaps make the error handling experience richer with little overhead.

jekirl:
I would propose that panics be only for truly unrecoverable errors, which in this case should be limited to debug mode perhaps? In general I think there should be no panics for the Mul/Add/Div/Sub/etc traits, and instead there be a checked_mul/div/add/sub/etc function for each implemented under the Decimal trait that returns an error type (through `error-chain`). Perhaps having panic on release and/or debug can be a feature flag?

paupino:
Have expanded `checked` operators into #80. I do like the idea of debug mode only panic's however would like to perhaps come up with a clear set of rules. For div, for example, I'm thinking:

- Div by zero = panic. If not panic, what would make sense to return? We could introduce the concept of `Infinity` within flags? 
- Overflow = Saturate
- Underflow = Round

Thoughts?

paupino:
Closing since this has been effectively handled now

PR #544, opened 2022-09-01
[v2] Replace error type with ones more like std

CAD97:
A start towards #49, though I haven't audited panicking, just migrated the existing errors.

Notably, because the error type no longer holds `String`, the crate no longer has a hard requirement on alloc.


paupino:
Thank you for this! I'll need to start a `v2` branch to start API breaking changes. There are a few things I'd ultimately like to address/clean up for `v2` while at the same time still be able to bug fix any smaller issues on the `v1` branch.

For the meantime, I'll leave this open and once I create a `v2` branch we'll see if we can retarget it and perform a full review. There are a couple of `v1` changes I'd like to get in before creating the branch (just to make it easier).

PR #547, opened 2022-09-06
Add parity scale codec

Gauthamastro:
PR adds support of `parity-scale-codec` for rust-decimal.

References
1. https://docs.substrate.io/reference/scale-codec/
2. https://github.com/paritytech/parity-scale-codec

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,8 @@ build = "build.rs"
 all-features = true
 
 [dependencies]
+parity-scale-codec = {package = "parity-scale-codec", optional = true, version = "3.1.5"}
+parity-scale-codec-derive = {package = "parity-scale-codec-derive", optional = true, version = "3.1.3"}
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
 borsh = { default-features = false, optional = true, version = "0.9.3" }
@@ -78,6 +80,7 @@ serde-with-float = ["serde"]
 serde-with-str = ["serde"]
 std = []
 tokio-pg = ["db-tokio-postgres"] # Backwards compatability
+scale-codec =  ["parity-scale-codec-derive","parity-scale-codec"]
 
 [workspace]
 members = [".", "./macros"]
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -30,6 +30,8 @@ use num_traits::float::FloatCore;
 use num_traits::{FromPrimitive, Num, One, Signed, ToPrimitive, Zero};
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
+#[cfg(feature = "scale-codec")]
+use parity_scale_codec_derive::{Decode,Encode};
 
 /// The smallest value that can be represented by this decimal type.
 const MIN: Decimal = Decimal {
@@ -124,6 +126,10 @@ pub struct UnpackedDecimal {
     archive_attr(derive(Clone, Copy, Debug))
 )]
 #[cfg_attr(feature = "rkyv-safe", archive_attr(derive(CheckBytes)))]
+#[cfg_attr(
+feature = "scale-codec",
+derive(Decode, Encode),
+)]
 pub struct Decimal {
     // Bits 0-15: unused
     // Bits 16-23: Contains "e", a value between 0-28 that indicates the scale


src/decimal.rs
```diff
@@ -548,7 +556,7 @@ impl Decimal {
     }
 
     #[must_use]
-    pub(crate) const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
+    pub const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
```
paupino, line 559: Was there a reason for changing the visibility of this? The reason it was kept this way as it allows flags to get into an invalid state. `from_parts` is similar, however ensures internal state is kept valid.

Cargo.toml
```diff
@@ -22,6 +22,9 @@ build = "build.rs"
 all-features = true
 
 [dependencies]
+parity-scale-codec = {package = "parity-scale-codec", optional = true, version = "3.1.5", features = ["max-encoded-len"], default-features = false}
```
paupino, line 25: Nit: could you make these alphabetical in the list? Also - we're making the attributes alphabetical too.
Btw, you shouldn't need to specify the `package` attribute since it is the same name as the crate.

src/decimal.rs
```diff
@@ -124,6 +128,10 @@ pub struct UnpackedDecimal {
     archive_attr(derive(Clone, Copy, Debug))
 )]
 #[cfg_attr(feature = "rkyv-safe", archive_attr(derive(CheckBytes)))]
+#[cfg_attr(
+feature = "scale-codec",
+derive(Decode, Encode, TypeInfo, MaxEncodedLen),
```
paupino, line 133: Side note (no change required): these `derive` attributes are causing a bit of code bloat. I want to rethink how we do this in v2. It may be by exposing a wrapper `struct` and aliasing it however all in all, this current approach isn't entirely scalable.

paupino (changes requested):
Thank you for your contribution! Sorry it's taken a while to look at this. Life - yada yada. 

A couple of changes requested as well as the need to run `makers format` to pass compilation.

Gauthamastro:
@paupino I got busy with my work and lost this PR through the cracks. We use your library extensively, I would be great if we can merge this PR; I will fix your comments soon.

src/decimal.rs
Gauthamastro: I agree it should be private; I probably forgot to change it after debugging.

Cargo.toml
Gauthamastro: Resolved

Cargo.toml
```diff
@@ -78,6 +81,7 @@ rkyv = ["dep:rkyv"]
 rkyv-safe = ["dep:bytecheck", "rkyv/validation"]
 rocket-traits = ["dep:rocket"]
 rust-fuzz = ["dep:arbitrary"]
+scale-codec =  ["parity-scale-codec-derive","parity-scale-codec","scale-info"]
```
paupino, line 84: We're starting to use the updated format for feature dependencies:
```suggestion
scale-codec =  ["dep:parity-scale-codec-derive","dep:parity-scale-codec","dep:scale-info"]
```

Cargo.toml
```diff
@@ -86,7 +90,7 @@ serde-str = ["serde-with-str"]
 serde-with-arbitrary-precision = ["serde", "serde_json/arbitrary_precision", "serde_json/std"]
 serde-with-float = ["serde"]
 serde-with-str = ["serde"]
-std = ["arrayvec/std", "borsh?/std", "bytecheck?/std", "byteorder?/std", "bytes?/std", "rand?/std", "rkyv?/std", "serde?/std", "serde_json?/std"]
+std = ["arrayvec/std", "borsh?/std", "bytecheck?/std", "byteorder?/std", "bytes?/std", "rand?/std", "rkyv?/std", "serde?/std", "serde_json?/std", "parity-scale-codec?/std","scale-info?/std"]
```
paupino, line 93: Just a simple reordering of inclusions:
```suggestion
std = ["arrayvec/std", "borsh?/std", "bytecheck?/std", "byteorder?/std", "bytes?/std", "parity-scale-codec?/std", "rand?/std", "rkyv?/std", "scale-info?/std", "serde?/std", "serde_json?/std"]
```

src/decimal.rs
```diff
@@ -120,6 +124,11 @@ pub struct UnpackedDecimal {
     archive(compare(PartialEq)),
     archive_attr(derive(Clone, Copy, Debug))
 )]
+#[cfg_attr(feature = "rkyv-safe", archive_attr(derive(CheckBytes)))]
```
paupino, line 127: This is duplicated below
```suggestion
```

src/decimal.rs
```diff
@@ -548,7 +557,7 @@ impl Decimal {
     }
 
     #[must_use]
-    pub(crate) const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
+    pub const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
```
paupino, line 560: Reverting this line:
```suggestion
    pub(crate) const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
```

paupino (changes requested):
Sorry it's taken a little while to review this!

Added a few changes as suggestions in the PR to make it a little easier.
One thing I noticed: there is no test covering this or documentation for this feature.

As an example PR for adding a test and docs, please see this: https://github.com/paupino/rust-decimal/pull/583/files

Almost there!

paupino:
I'm closing this as part of a clean up, namely since it is stale. That said, I'm not sure about creating a direct dependency (I'd like to revisit how we "connect" libraries such as these). There is of course a workaround by leveraging a transparent type right now (though I'd like to make this easier).

PR #551, opened 2022-09-21
Use the new feature resolver

c410-f3r:
# Previous  behaviour

The usage of an optional feature always pulls the optional dependencies listed in the same feature even if these dependencies weren't explicitly required, which is generally undesired, surprising and slower.

```toml
[dependencies]
rgb = { default-features = false, optional = true, version = "0.8" }

[features]
# `with-serde` will always bring `serde` as well as `rgb`
with-serde = ["rgb/serde", "serde"]
```

# Current behaviour

With the new "weak dependency" feature, optional dependencies won't be automatically pulled effectively solving the above problem.

```toml
[dependencies]
rgb = { default-features = false, optional = true, version = "0.8" }

[features]
# `rgb` won't be included unless the `rgb` feature is required
with-serde = ["rgb?/serde", "serde"]
```

# Future behaviour

The "name-spaced dependency" feature enables the listing of optional features using the name of any optional dependency, so, instead of creating features with prefixes, suffices or funny names to avoid overlapping, it will be possible to group related stuff in intuitive names.

```toml
[dependencies]
rgb = { default-features = false, optional = true, version = "0.8" }

[features]
# `serde` can now be used as a feature with `dep:`
serde = ["rgb?/serde", "dep:serde"]
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -192,7 +192,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.56.0
+          toolchain: 1.60.0
           profile: minimal
           override: true
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ path = "benches/comparison.rs"
 
 [package]
 authors = ["Paul Mason <paul@form1.co.nz>"]
+build = "build.rs"
 categories = ["science","mathematics","data-structures"]
 description = "Decimal number implementation written in pure Rust suitable for financial and fixed-precision calculations."
 documentation = "https://docs.rs/rust_decimal/"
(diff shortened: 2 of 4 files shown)


c410-f3r:
I personally prefer to declare all optional dependencies with `dep:` and then `#[features] foo = ["dep:foo"]` to avoid confusing with the older resolver but in this PR only the minimal has been modified.

paupino:
This looks good. Because it relies on `1.60` I want to make sure I hold off on releasing an entirely new version until `1.64` is released though that should be just around the corner given the 6 week release cycles - perhaps by tomorrow?

paupino:
I also like the explicit `dep:` syntax too! It has been a real pain trying to avoid feature/dependency conflicts to date 😂

paupino:
Also, thank you @c410-f3r once again for your help with this library!

c410-f3r:
Yeap, today is the day for 1.64.

Thank you @paupino for accepting PRs, maintaining this crate and issuing releases in a regular basis. These efforts are commendable because it is not rare to see open source projects that get stuck over time.

PR #553, opened 2022-09-25
Fix roundtrip through csv

yongqli:
Currently, a roundtrip through `csv` loses precision, even with the `serde-str` feature flag because `OptionDecimalVisitor` calls `deserialize_any()`. This can be easily fixed by taking advantage of the fact that Decimal's `serde::Deserialize` impl already has several versions depending on feature flags.

Tested with `cargo test --features serde --features serde-str`

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -390,7 +390,7 @@ impl<'de> serde::de::Visitor<'de> for OptionDecimalVisitor {
     where
         D: serde::de::Deserializer<'de>,
     {
-        d.deserialize_any(DecimalVisitor).map(Some)
+        <Decimal as serde::Deserialize>::deserialize(d).map(Some)
     }
 }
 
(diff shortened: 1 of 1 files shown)


paupino:
Unfortunately, this gets a little more complex once you start adding the matrix of features. As an example:

```
cargo test --features=serde-str,serde-float serde
```

You can curb against this a little by feature flagging the optional deserialization:

```rust
    #[cfg(not(feature = "serde-float"))]
    fn visit_some<D>(self, d: D) -> Result<Option<Decimal>, D::Error>
    where
        D: serde::de::Deserializer<'de>,
    {
        <Decimal as serde::Deserialize>::deserialize(d).map(Some)
    }

    #[cfg(feature = "serde-float")]
    fn visit_some<D>(self, d: D) -> Result<Option<Decimal>, D::Error>
    where
        D: serde::de::Deserializer<'de>,
    {
        d.deserialize_any(DecimalVisitor).map(Some)
    }
```

However, this only takes you part of the way. I still found that the roundtrip will still not work perfectly even in some simple cases when checking for:

```rust
assert_eq!(original.value.unwrap().unpack(), deserialized.value.unwrap().unpack());
```

giving:
```
---- serde::test::csv_roundtrip stdout ----
thread 'serde::test::csv_roundtrip' panicked at 'assertion failed: `(left == right)`
  left: `UnpackedDecimal { negative: false, scale: 3, hi: 0, mid: 0, lo: 123400 }`,
 right: `UnpackedDecimal { negative: false, scale: 1, hi: 0, mid: 0, lo: 1234 }`', src/serde.rs:867:9
```

All in all, the `serde` feature needs a whole lot of cleanup as it quite easily gets confusing. 

WIP for changes here: https://github.com/paupino/rust-decimal/compare/f/serde-csv-roundtrip?expand=1

paupino:
Since the fix that this PR issued is in master now am closing this PR. We can certainly reopen it if it gets rebased off of master and adds any additional features/tests.

PR #555, opened 2022-10-11
Support precision in `fmt_scientific_notation`

nicksenger:
Currently the precision field is ignored when formatting a `Decimal` in scientific notation. This PR updates `fmt_scientific_notation` to respect the precision if specified.


paupino:
Sorry for the delay in reviewing this!

Thank you for your help here. All in all, I like the direction however want to be a bit more deliberate with regards to when `precision` is defined vs not. What I'll do is I'll merge this then do a small refactor in a follow up PR to flip it around a bit. All in all, it'll be more verbose however I think it is a bit easier to follow rather than keeping track of `Some(precision)` vs `None` etc.

Again, sorry for the delay and thank you for your contribution!

nicksenger:
Sounds good, thank you!

PR #556, opened 2022-10-17
Move `size_32` rkyv feature to test dependencies.

jenanwise:
rkyv requires users to specify one of its `size_*` features. However, they are mutually exclusive, so if a library like `rust-decimal` enables `size_32`, end users will not be able to use e.g. `size_64`.

This is a breaking change for existing `rust-decimal` users who use the `rkyv` feature, as they will now need to explicitly set a `size_*` feature on their `rkyv` dependency.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,7 +63,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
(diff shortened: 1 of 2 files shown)


jenanwise:
Sadly I think the `size_*` approach that `rkyv` is using doesn't quite work, as exemplified by breaking `--all-features`. I'm going to follow up with that project and close this for now.

PR #560, opened 2022-12-05
Implement `TryFrom<&str>` for `Decimal`

c410-f3r:
Has the same implementation of `FromStr for Decimal`. 

Advantageous because `TryFrom` is already included in the prelude meaning that no import is necessary.

```rust
use core::str::FromStr;
let _ = Decimal::from_str("1");
```

```rust
let _ = Decimal::try_from("1");
```


paupino:
Thank you for your help with this library @c410-f3r !

c410-f3r:
Thank you @paupino !

PR #568, opened 2023-01-31
chore: bump `borsh` and `criterion`

attila-lin:
clear `dep:`


attila-lin:
fixed.

paupino:
Thank you for bumping these dependencies!

PR #575, opened 2023-02-09
fix: decimal round_dp on zero with too large dp

jon-chuang:
Let us start with regression test and work our way to a fix

(There is no fix, but we should panic early rather than silently error)


jon-chuang:
Actually, that was not the correct fix. We don't even have to panic.

paupino:
Thank you for fixing this - it's a good catch.

PR #576, opened 2023-02-28
Introduce `trunc_with_scale`

c410-f3r:
Feel free to close this PR if such thing is not desired.

It took me quite some time to figure out how to truncate the fractional part to a certain scale so I thought that other people could also benefit from this method.
The internal implementation is shamelessly re-using `round_dp_with_strategy` but in the future a more efficient code can be elaborated.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1050,6 +1050,21 @@ impl Decimal {
         }
     }
 
+    /// Returns a new `Decimal` with the fractional portion delimited by `scale`.
+    ///
+    /// # Example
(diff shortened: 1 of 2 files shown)


paupino:
Sorry for the delay - been very much busy! 
Happy to introduce this function if it makes discoverability easier. What I may do however is take what you have and update it so that this and `trunc` uses the same code path - it's a very minor change to be able to do so.

c410-f3r:
Nice! Thank you very much @paupino

PR #577, opened 2023-02-28
Enable support for scalar operations on ndarray

lmsxyz:
# Problem

I am currently using [ndarray](https://github.com/rust-ndarray/ndarray) with `rust-decimal`, but some operations requires the underlying structure to implement `ScalarOperation` marker.

# This PR

Enable such implementation under `ndarray` feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ Alternatively, you can edit your `Cargo.toml` directly and run `cargo update`:
 
 ```toml
 [dependencies]
-rust_decimal = "1.28"
-rust_decimal_macros = "1.28"
+rust_decimal = "1.29"
+rust_decimal_macros = "1.29"
 ```
 
 ## Usage
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Version History
 
+## 1.29.0
+
+### Added
+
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -168,6 +169,10 @@ Please note that `ln` and `log10` will panic on invalid input with `checked_ln`
 to curb against this. When the `maths` feature was first developed the library would return `0` on invalid input. To re-enable this
 non-panicking behavior, please use the feature: `maths-nopanic`.
 
+### `ndarray`
+
+Enables arithmetic operations using `ndarray` on arrays of `Decimal`.
```
paupino, line 174:
```suggestion
Enables arithmetic operations using [`ndarray`](https://github.com/rust-ndarray/ndarray) on arrays of `Decimal`.
```

paupino (changes requested):
Thanks for the contribution, and sorry for the delay getting around to a review! 
Overall - this looks great. I have one request: would you be able to add a test for this? Even though this doesn't look incredibly complex, it just helps make sure that an unrelated change doesn't break things for this feature.
A recent example of a feature with a test added is #520.
Thank you again!

paupino:
Just FYI (no pressure!): I'll look at releasing a new version of Rust Decimal later this week. If we can get a test wrapped up by then then I'll make sure this feature is included.

PR #582, opened 2023-03-16
add proptest feature for using arbitrary Decimals

cardoe:
Added support for generating arbitrary Decimals for proptest based tests which are feature gated by the proptest feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ diesel2 = { default-features = false, optional = true, package = "diesel", versi
 ndarray = { default-features = false, optional = true, version = "0.15.6" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
+proptest = { default-features = false, optional = true, features = ["std"], version = "1.0" }
 rand = { default-features = false, optional = true, version = "0.8" }
 rkyv = { default-features = false, features = ["size_32", "std"], optional = true, version = "0.7" }
 rocket = { default-features = false, optional = true, version = "0.5.0-rc.1" }
@@ -71,6 +72,7 @@ legacy-ops = []
 maths = []
 maths-nopanic = ["maths"]
 ndarray = ["dep:ndarray"]
+proptest = ["dep:proptest"]
 rand = ["dep:rand"]
 rkyv = ["dep:rkyv"]
 rkyv-safe = ["dep:bytecheck", "rkyv/validation"]
--- a/tests/decimal_tests.rs
+++ b/tests/decimal_tests.rs
@@ -4624,6 +4624,19 @@ mod generated {
     gen_test!(test_sub_111_111, "Sub_111_111.csv", checked_sub);
 }
 
+#[cfg(feature = "proptest")]
+mod proptest {
+    use super::Decimal;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn test_proptest_validate_arbitrary_decimals(num in any::<Decimal>()) {
+            assert!(num.is_zero() || !num.is_zero());
+        }
+    }
+}
+
 #[cfg(feature = "rocket-traits")]
 mod rocket {
     use crate::Decimal;
--- a/make/tests/misc.toml
+++ b/make/tests/misc.toml
@@ -1,11 +1,16 @@
 [tasks.test-misc]
 dependencies = [
+    "test-proptest",
     "test-rust-fuzz",
     "test-rocket-traits",
     "test-borsh",
     "test-rkyv",
     "test-rand"
 ]
+
+[tasks.test-proptest]
+command = "cargo"
+args = ["test", "--workspace", "--no-default-features", "--features=proptest", "--", "--skip", "generated"]
 
 [tasks.test-rust-fuzz]
 command = "cargo"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,8 @@ mod mysql;
     feature = "db-diesel2-postgres",
 ))]
 mod postgres;
+#[cfg(feature = "proptest")]
+mod proptest;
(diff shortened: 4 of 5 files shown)


Cargo.toml
```diff
@@ -33,6 +33,8 @@ diesel1 = { default-features = false, optional = true, package = "diesel", versi
 diesel2 = { default-features = false, optional = true, package = "diesel", version = "2.0" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
+proptest = { default-features = false, optional = true, features = ["std"], version = "1.0" }
```
paupino, line 36: Is `std` required? You could potentially add this to the `std` feature of Rust Decimal - i.e. `"proptest?/std"`

tests/decimal_tests.rs
```diff
@@ -4572,6 +4572,19 @@ mod generated {
     gen_test!(test_sub_111_111, "Sub_111_111.csv", checked_sub);
 }
 
+#[cfg(feature = "proptest")]
```
paupino, line 4575: Thanks for adding a test! Could you add it to the build too? Probably inside the `misc.toml` file: https://github.com/paupino/rust-decimal/blob/master/make/tests/misc.toml

paupino (changes requested):
Thank you for the contribution!

A couple of minor suggestions - for the `std` one it'd be good to move that into rust decimal's std feature just to avoid accidentally including `std` for `core` builds.

The last request would be to add something to the `README.md` just to improve visibility, if possible.

Thanks again!

Cargo.toml
cardoe: Unfortunately without it I get this error.

```
error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
   --> /Users/cardoe/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proptest-1.1.0/src/collection.rs:219:23
    |
219 |     [] fn VecToDeque[<T : fmt::Debug>](vec: Vec<T>) -> VecDeque<T> {
    |                       ^ unconstrained type parameter
```

I'm compiling rust-decimal with `cargo test --no-default-features --features proptest`

It's possible I'm missing something else or proptest has an issue?

tests/decimal_tests.rs
cardoe: Added. Thanks for pointing it out.

Cargo.toml
cardoe: So proptest requires an unstable compiler to work with no-std per their docs at https://altsysrq.github.io/proptest-book/proptest/no-std.html and they have an opened issue at proptest-rs/proptest#47 for resolving this. So can we leave the dependency on proptest with std?

paupino: Ok, interesting. While we can force `std` within the rust decimal feature it certainly highlights a flaw in the ecosystem - which isn't unique to `proptest`. 

Rust Decimal is really a small data type within the rust ecosystem providing not much more than a type to calculate decimal numbers. To provide support for various frameworks it is exposing features which provide compatibility. While exposing features for compatibility isn't a fundamental problem, having rust decimal pull in the dependencies etc of the peering library is. _Ideally_ if rust decimal was to provide the compatibility layer, it'd only need to pull in the smallest surface area of the library as possible. For `proptest` it'd really just be the `Arbitrary` trait and dependent types. That, or the inverse relationship should be exposed (i.e. `proptest` expose a rust decimal feature). The issue with the current approach is that we force `proptest` support to be `std` - whereas it shouldn't need to be. Rust Decimal certainly doesn't require `std` so we shouldn't really force a downstream dependency to be that either.

Again, this isn't unique to this PR/library - it's just a comment on the eco-system. There is already a precedent for rust decimal to provide the compatibility layer however it is perhaps something that I may need to rethink for v2. It's not an easy thing to solve as the intention is not to have a "stand off" with other libraries, however rather provide the most intuitive and flexible solution for the end user. Anyway!

Cargo.toml
```diff
@@ -34,6 +34,8 @@ diesel2 = { default-features = false, optional = true, package = "diesel", versi
 ndarray = { default-features = false, optional = true, version = "0.15.6" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres = { default-features = false, optional = true, version = "0.19" }
+proptest = { default-features = false, optional = true, features = ["std"], version = "1.0" }
+proptest-derive = { default-features = false, optional = true, version = "0.3" }
```
paupino, line 38: Do you need proptest-derive? I don't believe it's used in this implementation.

paupino: Just reading about that library, I'm pretty sure that it's probably unneeded and can be removed! :) It looks like you've implemented `Arbitrary` by hand so need for this too.

paupino:
```suggestion
```

Cargo.toml
```diff
@@ -71,6 +73,7 @@ legacy-ops = []
 maths = []
 maths-nopanic = ["maths"]
 ndarray = ["dep:ndarray"]
+proptest = ["dep:proptest", "dep:proptest-derive"]
```
paupino, line 76:
```suggestion
proptest = ["dep:proptest"]
```

Cargo.toml
cardoe: I think it would be great if there was a compatible Arbitrary trait between proptest, rust-fuzz and other testing libraries. None of them are compatible with each other unfortunately.

I totally agree with you about this issue in the eco-system. There's a couple of other areas in Rust that are wart-y in my opinion as well. Like MSRV support.

Cargo.toml
cardoe: No I don't. I originally was gonna use it but wrote the implementation by hand and didn't remove the depend.

paupino (approved):

PR #597, opened 2023-07-13
updated rkyv with reexported bytecheck

mkatychev:
This upstream PR removes `bytcheck` as a direct dependency for `rkyv/validation`:
https://github.com/rkyv/rkyv/pull/344
Thus it allows downstream crates such as this one to maintain one version for both `rkyv` & `bytecheck`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ all-features = true
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
 borsh = { default-features = false, optional = true, version = "0.10.0" }
-bytecheck = { default-features= false, optional = true, version = "0.6" }
 byteorder = { default-features = false, optional = true, version = "1.0" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -75,7 +74,7 @@ ndarray = ["dep:ndarray"]
 proptest = ["dep:proptest"]
 rand = ["dep:rand"]
 rkyv = ["dep:rkyv"]
-rkyv-safe = ["dep:bytecheck", "rkyv/validation"]
+rkyv-safe = ["rkyv/validation"]
```
mkatychev, line 77: Ideally this should be renamed to `rkyv-validation` to mirror the feature flag name upstream but that is a breaking change or there would have to be a `#[cfg(any("rkyv-safe", "rkyv-validation"))]` interim approach.

paupino: This is a good idea. While we need to keep `rkyv-safe` for backwards compatibility, it'd be easy to introduce a new feature with this name, particularly if it makes discoverability easier..

PR #603, opened 2023-08-18
Drop unnecessary byteorder dependency

psychon:
The byteorder dependency is only used to read u16 and i16 values in big endian from a slice of bytes. This can easily be done with just the standard library.


paupino:
Awesome - Thank you for your contribution!

PR #621, opened 2023-11-13
Upgrade Borsh to resolve vulnerability

paupino:
This replaces / closes #616 - Borsh now requires feature flags to be provided to enable the derive macros that this feature was utilizing.

Ultimately it resolves https://rustsec.org/advisories/RUSTSEC-2023-0033.html.


westy92:
Thank you so much! When can we expect a release?

paupino:
@westy92 Thanks for helping push this through - I'll prepare a release now and have something out shortly.

westy92:
@paupino you're welcome - and thank you!

PR #625, opened 2023-11-16
Fixes precision issue with to_f64 with some numbers without fractions

### LINKED ISSUE #624: To f64 losing accuracy for `100000.000000000000000000`
When `100000.000000000000000000` converting to f64, it will lose accuracy.
But for `10000.000000000000000000` and `1000000.000000000000000000`, it works well.

The test code is 
```
#[test]
fn test_decimal() {
    let decimal = Decimal::from_str_exact("100000.000000000000000000").unwrap();
    dbg!(f64::try_from(decimal).unwrap());
    let decimal = Decimal::from_str_exact("1000000.000000000000000000").unwrap();
    dbg!(f64::try_from(decimal).unwrap());
    let decimal = Decimal::from_str_exact("10000.000000000000000000").unwrap();
    dbg!(f64::try_from(decimal).unwrap());
}
```

The result is

image

The rust-decimal version is `1.32.0`

paupino:
Fixes #624 

When performing a conversion from `Decimal` to `f64` we would perform a round as the last step to remove any approximations that may have crept in during conversion. This works fairly well when there is a fractional component however is completely unnecessary when the number only has an integral component.

This removes the need to round if dealing with no fractional part.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2354,7 +2354,7 @@ impl ToPrimitive for Decimal {
             let integer = self.to_i128();
             integer.map(|i| i as f64)
         } else {
-            let sign: f64 = if self.is_sign_negative() { -1.0 } else { 1.0 };
+            let neg = self.is_sign_negative();
             let mut mantissa: u128 = self.lo.into();
             mantissa |= (self.mid as u128) << 32;
             mantissa |= (self.hi as u128) << 64;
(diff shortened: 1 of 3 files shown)


paupino:
Hmm - this is an interesting side effect. Will need to look into this. There is a test failing when both `serde-str` and `serde-float` are enabled (which is not recommended for use with `bincode`). 
I have a longer term approach for richer (de)serialization support via policies, however for the meantime this is unfortunately a supported combination I'll need to investigate.

> If, for some reason, you also have `serde-float` enabled then this will use `deserialize_f64` as a type hint. Because converting to `f64` loses precision, it's highly recommended that you do NOT enable this feature when working with `bincode`. That being said, this will only use 8 bytes so is slightly more efficient in terms of storage size.

PR #635, opened 2024-01-10
support bytemuck

dovahcrow:
One drawback for this PR is that it lowers the `unsafe_code` lint from forbid to deny because bytemuck requires unsafe.


robjtede:
Would #[cfg_attr(not(feature = "bytemuck), forbid(unsafe))]` work?

dovahcrow:
> Would `#[cfg_attr(not(feature = "bytemuck), forbid(unsafe))]` work?

Yeah this should work. But is it too verbose because you also need `#[cfg_attr(feature = "bytemuck), deny(unsafe)]`? I just feel deny is sufficient since it behaviours same as forbid but allows overridden.

robjtede:
Signalling forbid is important for consumers who run cargo-geiger and have no need of the bytemuck integration.

dovahcrow:
@robjtede done. I still leave deny/allow there since unsafe code is still not recommended.

PR #637, opened 2024-01-12
Expose mutually exclusive rkyv features for size.

gz:
rkyv by default serializes `usize` as `u32`. This isn't ideal on most modern platforms and unfortunately is configured by a set of (mutually exclusive) feature flags.

Currently rust-decimal sets it to 32-bit in it's `Cargo.toml` file. This means downstream apps can't override it (or just enable 64-bit de-serialization) as it will break compilation because these flags are mutually exclusive.

For libraries it's recommended by the rkyv maintainer to just use `default-features=false` and then "re-export" features so the user can control it the right serialization size instead of setting one. So this fix removes the features that were unnecessarily set by the Cargo.toml and adds some new features to activate different rkyv sizes. The approach is similar to what the ordered-float or chrono crate do:

https://github.com/reem/rust-ordered-float/blob/8111b345372632893af0b8aa12152f3dc7278aba/Cargo.toml#L37
https://github.com/chronotope/chrono/blob/6ec8f97d16ce320a6f948b1cd1494ed3a21b251f/Cargo.toml#L32


paupino (commented):
While this change seems simple, it actually causes a couple of issues:

* This breaks library upgraders - to avoid this you'd need to keep the default `rkyv` feature and point it to `rkyv-32`. To do this you'll also need to use the `dep:` format.
* Introducing these features will break documentation generation. We compile documentation using `--all-features` following the pattern defined [here](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

As an aside, it looks like the `main` branch of `rkyv` recently merged a change that resolves this feature exclusivity here: https://github.com/rkyv/rkyv/pull/461. Should we instead wait for this version to be released and remove the size element altogether?

paupino:
Closing as stale — the upstream rkyv fix has since landed, making this approach unnecessary. Thanks for the contribution @gz!

PR #638, opened 2024-01-13
Reimplement pow function for integer exponent.

### LINKED ISSUE #631: checked_powd takes exponentially long time to complete
Raising a high precision number to a vary large power blows up exponentially.

For example:

```rust
0.99999999999999 ** 14772894
```

takes a long time to complete.  Adding one more zero to the exponent and it takes forever.

Perhaps it should overflow or underflow somewhere within a sane limit...

schungx:
Fixes #631 and does not become pathological with larger exponents.


schungx:
For some reason, one of the tests come up with an off-by-one error at the last decimal place.

I'm not sure why but it does not show up on my computer, which passes all tests.

paupino:
It looks like it was failing when combined with the `legacy-ops` feature. At some stage I'll look at completing the deprecation and removing the feature (it's at least 4 years old past deprecation date). Anyway - I've gone ahead and removed the test for this feature for now.

Thank you for the PR!

PR #639, opened 2024-01-15
Implement traits on the ArchivedDecimal.

gz:
rkyv generates a separate type `ArchivedDecimal`, which represents the serialized value. In this PR we implement several traits to allow for ordering and equality on the `ArchivedDecimal` type (according to the rules of the Decimal type).

We also implement comparison between `Decimal` and `ArchivedDecimal` (PartialOrd) which is helpful to efficiently compare e.g., a more complex in-memory type that uses decimal as part of its fields with the serialized counter-part.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2586,7 +2586,18 @@ impl PartialEq for Decimal {
     }
 }
 
+#[cfg(feature = "rkyv")]
+impl PartialEq for ArchivedDecimal {
+    #[inline]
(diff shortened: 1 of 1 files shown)


gz:
hi, is there anything I can do to move this PR forward?

Tony-Samuels:
Closing as we're going to be stuck on `0.7` to avoid breaking changes (see #682), and we'll be pushing users to use the remote derive feature, where they can add the traits they need (see #683).

PR #644, opened 2024-02-02
Macros: remove circular dependency from macros feature

paupino:
This revisits the previous macro feature and removes the dependency on rust-decimal for parsing the string. Consequently, macros and rust decimal can be decoupled and delivered independently.

There may be a small regression in performance so I'm keeping this in draft as I diagnose.

New bench:
```
running 2 tests
test decimal_from_str            ... bench:          90 ns/iter (+/- 2)
test decimal_scientific_from_str ... bench:         265 ns/iter (+/- 8)
```

Old bench:
```
test decimal_from_str            ... bench:          69 ns/iter (+/- 1)
test decimal_scientific_from_str ... bench:         198 ns/iter (+/- 4)
```


paupino:
Closed due to it now being unnecessary

PR #646, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config


paupino:
At this point, since we don't check in a Cargo.lock there is not need for automatic updates of it. I think the primary goal should be to set minimal dependencies we are compatible with that avoids breaking the ecosystem. As an example, we still support diesel 1 as well as diesel 2. 

While I do appreciate the contribution, I'm going to close this for now to prevent additional noise. If perspectives change we can always reopen in the future.

PR #650, opened 2024-03-06
Add lightweight wasm support to `rust-decimal`

### LINKED ISSUE #613: Feature Request: Support wasm_bindgen for Decimal
I am using `rust_decimal` for precise calculations but unfortunately the following isn't supported
```rust
#[wasm_bindgen]
pub struct Precise {
    pub value: Decimal,
}
```
Therefore I want to request support for `wasm_bindgen` out of the box because support can't be implemented for foreign types.

ChristianIvicevic:
A new optional feature 'wasm' has been introduced in the codebase. This enables `wasm-bindgen` support, making `Decimal` compatible with the `wasm_bindgen` attribute macro, and exposes `fromNumber()` and `toNumber()` methods to convert between `Decimal` and the primitive `number` type.

Fixes #613

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,8 +393,9 @@ libraries, this crate can be compiled with `--no-default-features`.
 ### `wasm`
 
 Enable [`wasm-bindgen`](https://github.com/rustwasm/wasm-bindgen) support which makes `Decimal` compatible with the
-`wasm_bindgen` attribute macro and exposes `fromNumber()` and `toNumber()` methods to convert between `Decimal` and
-the primitive `number` type across boundaries.
+`wasm_bindgen` attribute macro and exposes the following methods across boundaries:
+* `fromNumber()` / `toNumber()` — convert between `Decimal` and the primitive `number` type.
+* `fromString()` / `toString()` — convert between `Decimal` and a string representation.
 
 ## Building
 
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -39,6 +39,9 @@ impl Decimal {
     }
 
     /// Returns the string representation of this `Decimal`.
+    ///
+    /// This intentionally overrides the default JS `toString()` so that string coercion
+    /// (e.g. template literals, `console.log`) produces the decimal representation.
     #[wasm_bindgen(js_name = toString)]
     #[must_use]
     pub fn to_string_js(&self) -> String {
--- a/tests/wasm.rs
+++ b/tests/wasm.rs
@@ -1,5 +1,9 @@
 #![cfg(target_arch = "wasm32")]
 
+// Tests use both `Decimal::from_str` (the standard `FromStr` trait) and `Decimal::from_string`
+// (the wasm-bindgen binding). This is intentional: `from_string` tests exercise the wasm API
+// surface, while `from_str` tests verify that core Decimal functionality works under wasm.
+
 use core::str::FromStr;
 use rust_decimal::Decimal;
 use wasm_bindgen_test::*;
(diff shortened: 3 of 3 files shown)


ChristianIvicevic:
I am not quite sure why the actions fail exactly but it feels like an issue with the macros crate referencing the published core crate which doesn't have the wasm support yet. Unfortunately I can't really figure out more than this. In the previous PR there were talks about having to adjust the CI most likely.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ rocket = { default-features = false, optional = true, version = "0.5.0-rc.3" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
+wasm-bindgen = { default-features = false, optional = true, version = "0.2" }
```
paupino, line 39: I'm just looking through how some other libraries introduce this and I think it needs to be done as a target as opposed to a simple feature. It could technically _still_ be a feature however it'd need to be a no-op for std/core.

e.g. 
```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-bindgen = { default-features = false, version = "0.2" }
```

There may be a requirement to also _exclude_ some other libraries that aren't supported on wasm of course. However that'd require some investigation (e.g. what features are supported in combination with wasm)

paupino (commented):
I think the way this is implemented in Cargo is the primary problem here - it needs to be target dependent as opposed to feature dependent. Currently, the build uses the `--all-features` flag which is likely where the issues stem (this is to ensure doc compilation works etc). Because WASM isn't an additive feature but moreso a target based feature we need to upgrade it as such.

For github actions, we also need to create another branch for testing WASM specifically. e.g. something like:
```yaml
      - name: Install rust
        uses: dtolnay/rust-toolchain@stable
        with:
          targets: wasm32-unknown-unknown

      - name: Check
        run: cargo check --target wasm32-unknown-unknown
```

I can help put this together over the next week or two.

ChristianIvicevic:
@paupino Were you able to set aside some time to investigate how to refactor the workflow in order to make this PR pass the pipeline?

.github/workflows/main.yml
```diff
@@ -124,12 +124,19 @@ jobs:
...
+          - beta / wasm
         include:
-          # This allows us to define targets in the future (e.g. wasm32-unknown-unknown)
           - name: stable
             rust: stable
           - name: beta
             rust: beta
+          - name: stable / wasm
+            rust: stable
+            target: wasm32-unknown-unknown
```
paupino, line 136: TODO: We need to make the tests adhere to this.

README.md
```diff
@@ -104,6 +104,7 @@ assert_eq!(total, dec!(27.26));
 * [rocket-traits](#rocket-traits)
 * [rust-fuzz](#rust-fuzz)
 * [std](#std)
+* [wasm](#wasm)
```
paupino, line 107: This is no longer a feature but can be it's own section - something along the lines of Wasm support and not all features support wasm. Currently supported = xyz.

.github/workflows/main.yml
paupino: (So the pass in the CI is not accurate)

Cargo.toml
```diff
@@ -37,19 +38,27 @@ serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = { default-features = false, version = "0.2" }
```
robjtede, line 42: It's a good idea to declare these things as optional **even with** the cfg guard, including `wasm-bindgen`/`js-sys`. Chrono found this out and reverted it: <https://github.com/chronotope/chrono/issues/1164>. There's some nasty circular dep situations that can occur otherwise.

paupino: Good call out and thank you for raising this @robjtede 

Admittedly, I'm still getting up to speed on this so it's good to find out what not to do also!

.github/workflows/main.yml
paupino: After looking a bit further into this, I think this is the wrong approach. I believe a wasm target needs to be handled completely separately.

ldicarlo:
Hello, thank you all for your work here! 
`rust_decimal` is a really useful crate. 
Do you expect to achieve this PR ? 
@ChristianIvicevic Would mind for some help? 
(not an expert here but anyway) 
@paupino would you care reviewing it once updated ? 

(not native english speaker, not trying to cause any harm here)
Thank you !

ldicarlo:
Still exploring it, but I think [tsify](https://github.com/madonoharu/tsify/?tab=readme-ov-file#type-override) does the job. 

```
#[derive(Tsify, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
#[tsify(into_wasm_abi, from_wasm_abi)]
pub struct SomeType {
    #[tsify(type = "number")]
    pub some_rust_decimal: Decimal,
}
```

Cargo.toml
```diff
@@ -38,22 +39,31 @@ serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = { default-features = false, version = "0.2", optional = true }
```
paupino: We could possibly tighten this, though for now it is probably ok

PR #656, opened 2024-03-28
Fix FromSql for Postgres Numeric NaNs

### LINKED ISSUE #655: Bug: Decimal::from_sql incorrectly transforms Postgres NaN to 0
Postgres Numerics have 3 special values: `NaN`, `Infinity`, and `-Infinity`. 

The `FromSql` implementation of `Decimal` currently returns an error when parsing Infinity and -Infinity because they have scales beyond what the crate accepts--this is fine, although it may be better to explicitly reject these values rather than relying on the scale, if they are not representable in `Decimal`.

But for NaN, `from_sql` silently converts it into a `0` because it is only checking if the `sign` is negative:

```
            neg: sign == 0x4000,
```

This is what the raw bytes of NaN look like in `from_sql`:
```
[/Users/luke/code/rust-decimal/src/postgres/driver.rs:68] &raw = Cursor {
    inner: [
        0,
        0,
        0,
        0,
        192,
        0,
        0,
        0,
    ],
    pos: 4,
}
[/Users/luke/code/rust-decimal/src/postgres/driver.rs:70] &sign = 49152
```

The relevant masks from postgres are:

```
#define NUMERIC_SIGN_MASK	0xC000
#define NUMERIC_POS		0x0000
#define NUMERIC_NEG		0x4000
#define NUMERIC_SHORT		0x8000
#define NUMERIC_SPECIAL		0xC000
#define NUMERIC_NAN		0xC000
#define NUMERIC_PINF		0xD000
#define NUMERIC_NINF		0xF000

```

My proposed logic for handling this is something like:
```
if sign & NUMERIC_SIGN_MASK) == NUMERIC_SPECIAL {
  if sign == NUMERIC_NAN || sign == NUMERIC_PINF || sign == NUMERIC_NINF {
     return some sort of Unsupported Error
  }
}
... continue parsing
```

If that logic sounds good I'd be happy to submit a patch

lukoktonos:
This fixes a bug where from_sql was converting Numeric::NaN to 0 rather than returning an error.

It also returns a more descriptive error message when from_sql is called with Numeric Infinity and -Infinity, which are also not representable in Decimal.

Closes #655

Side note: Unless I am missing something, it seems like the to_from_sql test wasn't actually being run since the `postgres` feature seems to no longer be a thing. I was able to get it to run by changing the cfg to `db-postgres`, so I added that to this commit as well.

### DIFF
--- a/tests/decimal_tests.rs
+++ b/tests/decimal_tests.rs
@@ -3516,7 +3516,7 @@ fn postgres_to_from_sql() {
 
 #[cfg(feature = "db-postgres")]
 #[test]
-fn from_sql_special_numeric() {
+fn postgres_from_sql_special_numeric() {
     use postgres::types::{FromSql, Kind, Type};
 
     // The numbers below are the big-endian equivalent of the NUMERIC_* masks for NAN, PINF, NINF
--- a/tests/decimal_tests.rs
+++ b/tests/decimal_tests.rs
@@ -3475,7 +3475,7 @@ fn declarative_ref_dec_sum() {
 
 #[cfg(feature = "db-postgres")]
 #[test]
-fn to_from_sql() {
+fn postgres_to_from_sql() {
     use bytes::BytesMut;
     use postgres::types::{FromSql, Kind, ToSql, Type};
 
--- a/tests/decimal_tests.rs
+++ b/tests/decimal_tests.rs
@@ -3473,7 +3473,7 @@ fn declarative_ref_dec_sum() {
     assert_eq!(sum, Decimal::from(45))
 }
 
-#[cfg(feature = "postgres")]
+#[cfg(feature = "db-postgres")]
 #[test]
 fn to_from_sql() {
     use bytes::BytesMut;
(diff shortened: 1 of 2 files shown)


tests/decimal_tests.rs
```diff
@@ -3473,7 +3473,7 @@ fn declarative_ref_dec_sum() {
     assert_eq!(sum, Decimal::from(45))
 }
 
-#[cfg(feature = "postgres")]
+#[cfg(feature = "db-postgres")]
 #[test]
 fn to_from_sql() {
```
paupino, line 3478: You're absolutely right about this - thank you. I think there is one more small change to ensure it is picked up by CI:

```suggestion
fn postgres_to_from_sql() {
```

tests/decimal_tests.rs
```diff
@@ -3514,6 +3514,36 @@ fn to_from_sql() {
     }
 }
 
+#[cfg(feature = "db-postgres")]
+#[test]
+fn from_sql_special_numeric() {
```
paupino, line 3519: Same thing here - (it filters postgres test by those that have postgres in the name):
```suggestion
fn postgres_from_sql_special_numeric() {
```

paupino (approved):
Thank you for your contribution! I'm going to rename those two functions so that they get included in CI but I think it should be all good to merge.

PR #657, opened 2024-04-10
Fix negation with overflow

benschulz:
Fixes an issue where `i128::MIN` would cause overflow.


paupino:
This is a great catch, thank you. The blindness in the test is a worry - I see the u128 test is also blind so I'll likely fix that as part of this too.

PR #659, opened 2024-07-07
Replace filtery with a strategy for implementation of proptest::Arbitrary

uzytkownik:
Currently the `proptest::Arbitrary` implementation uses `filter_map` with failure rate of ~90%. This causes problems if user need to filter it further.

This change replaces it with using a range removing backtracing.


paupino:
Thank you for this! 

For context for those looking into this - the previous implementation was generating a random number for scale however discarding most of them. This is quite inefficient - this updated implementation instead places a range on scale generation so that it can succeed without having to discard items.

PR #661, opened 2024-07-11
Bump macro version to 1.35

### LINKED ISSUE #660: rust_decimal_macros 1.35.0 not on crates.io (Or: Error in Documentation)
Newest version of `rust_decimal_macros` on crates.io is `1.34.2`, but the documentation explicitly uses `1.35`. 

As `rust_decimal_macros-1.34.2` seems to work fine with `rust_decimal-1.35.0` I see two solutions:

- Release `rust_decimal_macros-1.35.0` to sync versions
- Fix documentation to use `rust_decimal_macros = "1.35"` in `[dependencies]`

paupino:
Fixes #660

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rust_decimal_macros"
-version = "1.34.2"
+version = "1.35.0"
 authors = ["Paul Mason <paul@form1.co.nz>"]
 edition = "2021"
 description = "Shorthand macros to assist creating Decimal types."


paupino:
Hmm... Failure is from a dependent library here. Looking into this.

paupino:
It looks like the error may be coming from the mysql dependency that diesel1 is using. It seems like I may need to hard deprecate it - but just seeing if there is any alternative.

paupino:
This is a build only issue so I'm going to force this in and fix the dependecy issue separately

PR #663, opened 2024-07-11
Deprecate diesel1 feature and upgrade diesel to v2.2

### LINKED ISSUE #662: Diesel1 feature is breaking the build
Diesel1 is breaking the build with an old version of bindgen being linked.
We should consider either deprecating the feature or overloading the downstream dependency.

paupino:
Fixes #662 

This is a breaking change, however I think it is necessary to avoid being held back by this feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
 authors = ["Paul Mason <paul@form1.co.nz>"]
 build = "build.rs"
-categories = ["science","mathematics","data-structures"]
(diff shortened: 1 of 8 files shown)


paupino:
My reasoning for updating this is that if they have upgraded to the latest diesel then they're probably likely to also upgrade to the latest rust-decimal. Likewise, if they're using the older diesel, they're less likely to upgrade to the new rust-decimal.

So while it is breaking, my reasoning is that it is a permissible breaking change.

PR #674, opened 2024-08-28
Use repr(C) and add optional bytemuck support

kevinheavey:
Changes:

- Sets `repr(C)` on the `Decimal` struct
- Adds a `bytemuck` feature that derives `Pod` and `Zeroable` on `Decimal`


manglemix:
Is this still in progress? bytemuck is useful for me to do a safe union type

paupino:
I'm going to merge this into V1, however for future versions would love to see how we can conditional dependencies like this. I'll add it to the README and merge.

PR #678, opened 2024-09-18
Fix `f64` dec limit edge case

finnbear:
Right now the `f64`
```
79228162514264337593543950336
```
Converts to the decimal
```
79228162514264337593543950330
```

Under `Decimal::from_f64` or `Decimal::from_f64_retain`.

Which is wrong since:
1) the float overflowed and that's not the max decimal
2) conversion to decimal should have returned `None` anyway

The root cause is losing precision when dividing by 5, leading to missing an overflow when later multiplying by 10.

The included test (specifically the middle assertion) would fail without the patch.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2190,6 +2190,9 @@ fn base2_to_decimal(
             exponent10 -= 1;
             exponent5 += 1;
             ops::array::shl1_internal(bits, 0);
+        } else if exponent10 * 2 > -exponent5 {
+            // Multiplying by >=2 which, due to the previous condition, means an overflow.
+            return None;
         } else {
             // The mantissa would overflow if shifted. Therefore it should be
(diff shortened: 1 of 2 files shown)


paupino:
Thank you for this! 

I was just doing some playing around here and I believe there are some other edge cases too. 
As an example `Decimal::MAX` (as f64) and `Decimal::MAX - 1` will both also fail since the current max for the `f64` parsing algorithm is `79_228_162_514_264_337_593_543_950_330`. 

That said, I think this is taking things in the right direction so will merge this in.

finnbear:
@paupino AFAICT those problems don't exist since `f64` can't represent those values. My test covers three adjacent `f64` values, found via `f64::next_down`.

paupino:
@finnbear Oh cool - I wasn't aware of that function. A handy one to have in the toolkit - particularly for testing floats!

PR #679, opened 2024-09-19
[WIP] Feat: add bitcode support

l-7-l:
Hey, thank you for your amazing crate! This PR adds support for [bitcode](https://github.com/SoftbearStudios/bitcode).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
+bitcode = { version = "0", default-features = false, features = ["derive"], optional = true }
 borsh = { default-features = false, features = ["derive", "unstable__schema"], optional = true, version = "1.1.1" }
 bytes = { default-features = false, optional = true, version = "1.0" }
 diesel = { default-features = false, optional = true, version = "2.2.3" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -21,6 +21,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 arbitrary = { default-features = false, optional = true, version = "1.0" }
 arrayvec = { default-features = false, version = "0.7" }
+bitcode = { version = "0", default-features = false, features = ["derive"], optional = true }
```
l-7-l, line 24: [Add bitcode to libraries without specifying the major version so binary crates can pick the version.](https://github.com/SoftbearStudios/bitcode/blob/15fe3aa8393276c0312896478e2b81d4e28267da/README.md?plain=1#L40)

src/decimal.rs
```diff
@@ -112,6 +112,7 @@ pub struct UnpackedDecimal {
     archive_attr(derive(Clone, Copy, Debug))
 )]
 #[cfg_attr(feature = "rkyv-safe", archive(check_bytes))]
+#[cfg_attr(feature = "bitcode", derive(bitcode::Encode, bitcode::Decode))]
```
finnbear, line 115: Should also consider implementing manually to save 3 bytes per `Decimal`, or a ~18.75% size improvement. In particular, can store `flags` in `u8` instead of `u32` (because it only has 7 meaningful bits).

PR #685, opened 2024-10-15
Create pub constant Decimal::MAX_SCALE

### LINKED ISSUE #684: Expose `MAX_PRECISION` constant
Hey!

I've noticed the docs mention `Decimal::MAX_PRECISION`:

https://github.com/paupino/rust-decimal/blob/fa8857946cfb3bc4a08244b60133082d8dc575e7/src/decimal.rs#L873

However, this constant isn't currently exposed, and it is useful as it determines whether or not certain operations, such as `set_scale`, can error.

Could this constant be exposed? (And perhaps any other missing essential constants from the `constants` module?)

Thanks!

Tony-Samuels:
Resolves #684.

Changes are as follows:
- Change name from MAX_PRECISION to MAX_SCALE, to more accurately reflect what it does
- Switch (most) docs to refer to `Decimal::MAX_SCALE` instead of a raw number
- Switch internal usage to `Decimal::MAX_SCALE`, mostly just for consistency

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -525,7 +525,7 @@ impl Decimal {
     /// ```
     #[must_use]
     pub const fn from_parts(lo: u32, mid: u32, hi: u32, negative: bool, scale: u32) -> Decimal {
-        assert!(scale <= Self::MAX_SCALE, "Scale would exceed maximum support scale");
+        assert!(scale <= Self::MAX_SCALE, "Scale exceeds maximum supported scale");
         Decimal {
             lo,
             mid,
@@ -543,7 +543,7 @@ impl Decimal {
 
     #[must_use]
     pub(crate) const fn from_parts_raw(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
-        let decimal = if lo == 0 && mid == 0 && hi == 0 {
+        if lo == 0 && mid == 0 && hi == 0 {
             Decimal {
                 lo,
                 mid,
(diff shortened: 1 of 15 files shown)


src/decimal.rs
```diff
@@ -504,14 +513,7 @@ impl Decimal {
...
     /// * `hi` - The high 32 bits of a 96-bit integer.
     /// * `negative` - `true` to indicate a negative number.
-    /// * `scale` - A power of 10 ranging from 0 to 28.
-    ///
-    /// # Caution: Undefined behavior
-    ///
-    /// While a scale greater than 28 can be passed in, it will be automatically capped by this
-    /// function at the maximum precision. The library opts towards this functionality as opposed
-    /// to a panic to ensure that the function can be treated as constant. This may lead to
-    /// undefined behavior in downstream applications and should be treated with caution.
```
Tony-Samuels, line 514: Whilst reviewing docs for changes I also spotted that we can remove this undefined behaviour in `from_parts` and `from_parts_raw`. Given that we describe it as undefined behaviour previously, and a `panic` is a valid result (IMO) from undefined behaviour, it _could_ be argued this is not a breaking change. 

It's also worth noting that the documentation on what the function would do in this case (automatically capped by this function at the maximum precision) is not what it did anyway (it instead found the remainder past the end of the scale).

Note that the random generation functions needed fixing as they relied on this scale, and there was one additional test failure: `issue_384_neg_overflow_during_subtract_carry` hits the new `assert` for `from_parts`. I just calculated what the value would have been with the old behaviour (`3873892070 % 29 == 5`) to resolve this.

src/decimal.rs
```diff
@@ -549,7 +552,14 @@ impl Decimal {
             }
         } else {
             Decimal { flags, hi, lo, mid }
-        }
+        };
+
+        assert!(
```
paupino, line 557: Since this is a crate only function I'd perhaps remove this assert. Only because it can _sometimes_ be useful to generate invalid scales etc.

paupino (commented):
Thank you for taking the lead on this. A couple of small suggestions, but overall looks good.

Issue #688, opened 2024-10-24
Directly Exportable `dec!()` as Macro by Example


daniel-pfeiffer:
Since the companion macro does very little to justify a separate crate and proc-macro, I’ve sketched a solution with macro by example. For good measure I also propose `dec!(a123, radix)` and `dec!(123 >> scale)`.

I’ve done a little dummy `Decimal` with a dummy `const` parser (that calculates a checksum just for feasability.) Instead of iterating over chars, it manually loops over bytes. This is good enough, as we’re only parsing Ascii chars.

Surprisingly, `const fn`s aren’t evaluated at compile time – even when called with constant arguments. So as long as this wants to be compatible to pre-1.79 Rust (which introduced `const {}` for this) I’m assigning to a block local `const` as a workaround.

```rust
use core::fmt;

struct Decimal(&'static str, u8, u16);
impl Decimal {
    const fn from_str_or_panic(num: &'static str) -> Self {
        Self::from_str_radix_or_panic(num, 10)
    }
    const fn from_str_with_scale_or_panic(num: &'static str, scale: u32) -> Self {
        Self::from_str_radix_or_panic(num, (scale % 0xff) as u8) // todo
    }
    const fn from_str_radix_or_panic(num: &'static str, radix: u8) -> Self {
        let bytes = num.as_bytes();
        let mut i = num.len();
        let mut sum = 0;
        loop {
            if i == 0 { break }
            i -= 1;
            sum += match bytes[i] as char {
                '0'..='9' => bytes[i] - '0' as u8,
                'a'..='z' => bytes[i] - 'a' as u8 + 10,
                'A'..='Z' => bytes[i] - 'A' as u8 + 10,
                '.' => 37,
                '-' => 38,
                '_' => 0,
                ' ' => 0, // stringify!(-$num) inserts a space
                _ => { panic!("Malformed number") }
            } as u16;
        }
        Self(num, radix, sum)
    }
}

impl fmt::Display for Decimal {
    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(fmt, "{}#{} sum {}", self.1, self.0, self.2)
    }
}

// Simpler form from Rust 1.79 onwards in comments
macro_rules! dec {
    // While 1a is a literal, a1 is an ident. And, unlike literal, those don't take negation along.
    ($num:ident, $radix:literal) => {
        { const D: Decimal = Decimal::from_str_radix_or_panic(stringify!($num), $radix); D }
[...]

PR #692, opened 2024-12-10
Add `dec!()` macro by example

daniel-pfeiffer:
As [discussed](https://github.com/paupino/rust-decimal/issues/688), here is a cleaned up, further optimised and integrated new implementation of `dec!`. It is a completely `const` and compile-time drop in replacement. This requires Rust 1.79.

Your `Result<…String…>` is not destructible, i.e. not usable in `const`. Therefore the new parser has its own `Result`, that can be transparently converted by `?`. This also required splitting `try_from_i128_with_scale` which thus became simpler.

For now the parser does only exact in `i128`. Rounding should be addable, if you want to switch parsing completely to this.

Being generic on a smaller type doesn’t currently seem possible in `const`. Should a benchmark show this to be slower than parsing into `u64`, the old and new parsers might continue to coexist, till a solution is found.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,7 @@ mod constants;
 mod decimal;
 mod error;
 mod ops;
-mod str;
+pub mod str;
 
 // We purposely place this here for documentation ordering
 mod arithmetic_impls;
@@ -57,18 +57,15 @@ pub use error::Error;
 #[cfg(feature = "maths")]
 pub use maths::MathematicalOps;
 
-// #[cfg(feature = "macros")]
-// pub use rust_decimal_macros::dec;
-
 /// A convenience module appropriate for glob imports (`use rust_decimal::prelude::*;`).
 pub mod prelude {
     #[cfg(feature = "maths")]
     pub use crate::maths::MathematicalOps;
+    #[cfg(feature = "macro")]
+    pub use crate::dec;
     pub use crate::{Decimal, RoundingStrategy};
     pub use core::str::FromStr;
     pub use num_traits::{FromPrimitive, One, Signed, ToPrimitive, Zero};
-    // #[cfg(feature = "macros")]
-    // pub use rust_decimal_macros::dec;
 }
 
 #[cfg(feature = "diesel")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,10 +59,10 @@ pub use maths::MathematicalOps;
 
 /// A convenience module appropriate for glob imports (`use rust_decimal::prelude::*;`).
 pub mod prelude {
-    #[cfg(feature = "maths")]
-    pub use crate::maths::MathematicalOps;
     #[cfg(feature = "macro")]
     pub use crate::dec;
+    #[cfg(feature = "maths")]
+    pub use crate::maths::MathematicalOps;
     pub use crate::{Decimal, RoundingStrategy};
     pub use core::str::FromStr;
     pub use num_traits::{FromPrimitive, One, Signed, ToPrimitive, Zero};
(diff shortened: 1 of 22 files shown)


src/lib.rs
```diff
@@ -64,7 +64,7 @@ pub use maths::MathematicalOps;
 pub mod prelude {
     #[cfg(feature = "maths")]
     pub use crate::maths::MathematicalOps;
-    pub use crate::{Decimal, RoundingStrategy};
+    pub use crate::{dec, Decimal, RoundingStrategy};
```
paupino, line 67: If we feature flag it, we'd just need to optionally include it here.

src/str.rs
```diff
@@ -711,6 +711,299 @@ pub(crate) fn parse_str_radix_n(str: &str, radix: u32) -> Result<Decimal, Error>
...
+///
+/// Normally this macro is performed at compile time. Alas that only allows limited error messages. If you
+/// put this attribute before any other parameters, evaluation is deferred. This gives richer error messages,
+/// useful if you don’t understand why your input is not being accepted.
+///
+/// Furthermore the expressions you pass to the options, which must normally be `const`, become dynamic:
+/// `dec!(#![run_time] radix: my_radix, exp: my_exp, 10)`
+#[macro_export]
+macro_rules! dec {
+    (#![run_time] $($rest:tt)+) => {
```
paupino, line 742: Small nit: typically parameter attributes would not have the exclamation mark (e.g. `#[run_time]`. That said, I wonder if it's worthwhile just splitting this out completely - e.g. `dec_debug!(xyz)` or `debug_dec!(xyz)`

paupino (changes requested):
Sorry for the delay in getting to this - and thank you for putting this together!

Added a couple of comments, however it really boils down to two requested changes I think:

First is to put this behind a feature flag `macros`. That way, we can remove some unnecessary ambiguity (also, making this an opt in feature). It means prelude will keep working until people have a chance to migrate over.

The _other_ thing we should do is clean up the old `macros` folder (i.e. remove it completely). Personally, I think having an opinionated way forward is better - also, since it is a separate crate it should be fine to completely remove (i.e. folks can still use the old crate). Of course, this requires some clean up of the documentation to remove the reference to the crate too.

Thanks again!

src/str.rs
daniel-pfeiffer: This emulates an outer attribute, as though it had been written before the macro. Really making one that can be put outside would be a big effort. So I opted for putting it inside, which in Rust is written with an exclamation mark.

It's a little more than debug, also giving dynamic parameters, which is why I called it run_time.

For the feature gate, do you see that as only affecting the prelude, or the overall availability?

paupino: Sorry - I may have been a little ambiguous! I realize I also skipped over the comment "Furthermore the expressions you pass to the options, which must normally be `const`, become dynamic" which kind of rules out the renaming of the macro (e.g. to `dec_debug!`). 

Anyway, I wonder if we should remove the exclamation mark (`!`) so it matches the current parameter attribute style in Rust?

e.g.
```rust
dec!(#[run_time] radix: my_radix, exp: my_exp, 10);
```

Subtle, but effectively it's just removing the exclamation mark in the `run_time` attribute. e.g. see here for an example: https://rust-lang.github.io/rfcs/2565-formal-function-parameter-attributes.html

Regarding the feature gate - I think yes for prelude for sure. Ideally it'd be the macro too (e.g. wrap in a `mod` with a conditional use perhaps?) however if it's too tricky then it's probably not that problematic to keep it.

daniel-pfeiffer: For the exclamation mark I don’t care, it’s your call. In this case it doesn’t matter, because it’s only emulating an attribute.

However you don’t seem to realise its meaning. Here are two equivalent ways of adding an attribute. When putting it inside the block (or a module…) it pertains to, Rust needs the exclamation mark. Else it would pertain to one next item.

```rust
#[doc = "same as /// as an outer attribute\n"]
fn f() {
    #![doc = "same as ///, but as an inner attribute"]
}
```

paupino: I see what you mean - it pertains to all arguments in this case and changes the macro behavior itself. 

Ok, cool let's just keep it. Thanks for working through this!

paupino (approved):

paupino:
Thank you for your help!

Issue #697, opened 2025-01-07
Panic found during fuzzing


schungx:
Fuzzing the following when parsing with `from_scientific`:

`-20165.4676_e-+4294967292`

Error: https://github.com/paupino/rust-decimal/blob/master/src/decimal.rs#L589

`attempt to add with overflow`

I suspect a few other positions may also have overflow/underflow panics with `exp`.

PR #700, opened 2025-02-06
Fix panic when parsing improper scientific notation.

schungx:
This fixes https://github.com/paupino/rust-decimal/issues/697


tests/decimal_tests.rs
```diff
@@ -3373,10 +3373,18 @@ fn it_can_parse_scientific_notation() {
         ("12345E-28", "0.0000000000000000000000012345"),
         ("1.2345E0", "1.2345"),
         ("1E28", "10000000000000000000000000000"),
+        (
+            "-20165.4676_e-+4294967292",
+            "ERR:Scale exceeds the maximum precision allowed: 4294967292 > 28",
```
paupino, line 3378: Minor: Stylistically, I'd probably shift this test case to use a `Result` instead - just to keep "the way to write tests" consistent (e.g. like this: https://github.com/paupino/rust-decimal/blob/46fb4c3c517bc0c27cd534b65f9e8b57c24ba18e/tests/decimal_tests.rs#L3243-L3275)

paupino (approved):
Thank you for fixing this!

tests/decimal_tests.rs
schungx: Sorry for the late reply.  Pls go ahead and make this change.  Your style definitely look better.

Feel free to close this PR and redo it your way.

paupino: Let's merge your PR first so that you get credit for the contribution. We can always modify the style of the test afterwards (to be fair there are a few to make consistent).

PR #701, opened 2025-02-17
Upgrade rand crate version to 0.9

redforks:
I'm encountering a dependency conflict because I am using the `fake` crate along with `rust_decimal`, and the latest version of `fake` requires `rand` version 0.9.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ ndarray = { default-features = false, optional = true, version = "0.15.6" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres-types = { default-features = false, optional = true, version = "0.2" }
 proptest = { default-features = false, optional = true, features = ["std"], version = "1.0" }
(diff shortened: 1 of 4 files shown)


Tony-Samuels:
Same as #702, this would be a breaking change as is.

redforks:
Split the `rand` feature into two separate features: `rand_08` and `rand_09`, in order to support both versions.

I encountered a failure with `cargo test -F rand_09`. It appears that during unit testing, `OsRng` implements the 0.8
version of `RngCore`, but I need to access methods defined on the 0.9 version of the `Rng` trait. But I checked all cfg attribute, it should not happen, maybe something happened in rand 0.9.

In `cargo make` unit tests, I only enabled rand_08 unit tests.

Doc-tests are okay for either _08 or _09 features.

paupino:
Thank you for your contribution! I'm going to close this one since I instead merged #702 which is fixing the same thing. The only reason for picking the latter is because it remains backwards compatible (i.e. `rand` still points to 0.8, `rand-0_9` points to 0.9).

PR #702, opened 2025-02-17
feat: add rand-0_9 crate feature

robjtede:


Tony-Samuels:
This seems like a breaking change. Thoughts on making this a different feature and keeping the old version?

PR #708, opened 2025-03-07
Rust 2024

Billy-Sheppard:
Few things of note.

- I've fixed all the lints clippy requested, but I can revert if not desired
- I've updated all crates as new as possible (excl `bincode`)
- Passes all tests locally (except for `readme_deps`)
- Compiles locally (except for one error with diesel?)
- Had to convert the `#![forbid(unsafe_code)]` to a `#![deny(unsafe_code)]`
  - this is for:
```rust
#[allow(unsafe_code)]
#[cfg(feature = "rkyv")]
unsafe impl Portable for Decimal {}
```

Let me know how else I can confirm this is ready for a merge.


Tony-Samuels:
> This library maintains support for rust compiler versions that are 4 minor versions away from the current stable rust compiler version.

I haven't actually tested this to be certain, but I expect we'll need to hold off on a 2024 edition bump for another few minor releases before we can release this and uphold [our guarantees](https://github.com/paupino/rust-decimal?tab=readme-ov-file#minimum-rust-compiler-version).

Billy-Sheppard:
Excellent, no worries

Tony-Samuels:
This has been done on the master branch now, which will become the basis of the 2.0 release.

PR #711, opened 2025-03-17
Refactor macro for backwards compatability 

paupino:
This refactors the `dec` macro to fold in the benefits of the new behavior while keeping the benefits of the old behavior:

* Macros can now be included via a feature flag `macros` which exposes the `dec` macro both at the crate root, and within prelude.
* Converted new macros from `macro_rules` to compile time macros. This is important to maintain compile time errors (e.g. otherwise, they'd effectively panic during runtime evaluation).
* General linting cleanup and review

This also prepares the stage for 1.37.0.


macros/src/str.rs
```diff
@@ -0,0 +1,376 @@
...
+                    f,
+                    "Cannot parse decimal, unexpected \"{}\".",
+                    core::str::from_utf8(src).unwrap()
+                )
+            }
+        }
+    }
+}
+
+pub const fn parse_decimal(src: &str, exp: i32) -> ParseResult {
```
paupino, line 79: I think having `exp` as a mandatory argument is incorrect since it can get overwritten in `parse_10` cases.

macros/src/lib.rs
```diff
@@ -52,35 +73,33 @@ use rust_decimal::Decimal;
...
+        Err(e) => panic!("{}", e),
     };
 
-    let unpacked = decimal.unpack();
     expand(
-        unpacked.lo,
-        unpacked.mid,
-        unpacked.hi,
-        unpacked.negative,
+        unpacked.lo(),
```
paupino, line 99: Unnecessary since the i128 constructor is a const - that said, it doesn't hurt either.

paupino:
Comments added for future tasks - they don't affect the functionality but are future proofing.

PR #718, opened 2025-03-25
More efficient is_zero check

ivanrg99:
`rustc` is not smart enough to only use `or` for the generated assembly and adds extra instructions for jumps and compares, to facilitate the short-circuiting check. With a simple bit comparison we achieve the same effect and the compiler
does not generate any branching instructions:

https://godbolt.org/z/6nrhvsxs7

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -718,8 +718,9 @@ impl Decimal {
     /// assert!(num.is_zero());
     /// ```
     #[must_use]
+    #[inline]
     pub const fn is_zero(&self) -> bool {
-        self.lo == 0 && self.mid == 0 && self.hi == 0
+        self.lo | self.mid | self.hi == 0
     }
 
     /// Returns true if this Decimal number has zero fractional part (is equal to an integer)


Tony-Samuels:
Looks like we'll need a clean-up MR first to fix some issues not introduced by this change.

paupino:
Nice find!

I think this should target the v1 branch as opposed to master. It seems like a stable fix, which can also be applied to the currently unstable master. (I.e, merge/release against v1, then v1 propagates up to master).

ivanrg99:
> I think this should target the v1 branch as opposed to master. It seems like a stable fix, which can also be applied to the currently unstable master. (I.e, merge/release against v1, then v1 propagates up to master).

Should I create a new PR then with v1 as the base that targets v1 and close this one?

Tony-Samuels:
> Should I create a new PR then with v1 as the base that targets v1 and close this one?

I think we'd like this on both. If you leave this one open and create a new one for v1, then I'll make sure it's merged when we sort out master.

paupino:
Oh, yeah - it looks like you may need to rebase your change onto v1 first. 

Sorry, I've started using master for unstable breaking change cleanups (aka v2). I need to make sure it's clear for any new contributors what to do here - maybe a pull request template could help?

PR #720, opened 2025-03-31
Update to bincode v2

intendednull:
- [x] derive bincode::{Encode, Decode} for Decimal
- [ ] bincode v2 bumps msrv to 1.85
- [ ] tests are against bincode v1, which uses serde; i'd have to update the tests to use v2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ rocket = { default-features = false, optional = true, version = "0.5.0-rc.3" }
 serde = { default-features = false, optional = true, version = "1.0" }
 serde_json = { default-features = false, optional = true, version = "1.0" }
 tokio-postgres = { default-features = false, optional = true, version = "0.7" }
+bincode = { default-features = false, features = ["derive"], optional = true, version = "2.0.0-rc.3" }
 
 [dev-dependencies]
 bincode = { default-features = false, version = "1.0" }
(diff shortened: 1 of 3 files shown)


intendednull:
Looks like this code is failing when self.to_string() == "0". Not sure how that's possible tbh

```rust
#[cfg(all(feature = "serde-float", feature = "serde-arbitrary-precision"))]
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
            .serialize(serializer)
    }
}
```

src/serde.rs
```diff
@@ -675,17 +678,20 @@ mod test {
         ];
         for &raw in data.iter() {
             let value = Decimal::from_str(raw).unwrap();
-            let encoded = serialize(&value).unwrap();
-            let decoded: Decimal = deserialize(&encoded[..]).unwrap();
+            let encoded = encode_to_vec(&value, config::standard()).unwrap();
+            let decoded: Decimal = decode_from_slice(&encoded[..], config::standard()).unwrap().0;
             assert_eq!(value, decoded);
-            assert_eq!(8usize + raw.len(), encoded.len());
+            assert_eq!(1usize + raw.len(), encoded.len());
```
intendednull, line 684: Not sure this is correct. I check the str serialization path is running, but I'm fuzzy with the internal representation. Not sure why it would be smaller tbh

intendednull:
I've kept all the serde tests using the serde traits. Maybe we should add tests using the bincode Encode/Decode trats as well

paupino:
Closing as stale. Thanks for the work on this @intendednull!

For anyone picking this up, the outstanding items were:
1. `bincode_serialization_serde_float` fails with `serde-arbitrary-precision` enabled
2. MSRV bump to 1.85 needs documenting
3. Tests using bincode's native `Encode`/`Decode` traits should be added

PR #723, opened 2025-04-20
Ci fix

Tony-Samuels:
I _could_ have just done `r#gen()` to fix the old rand version, but I didn't see the point. If we're releasing a 2.0 version, then we can expect our users to be on relatively new versions of everything.


Cargo.toml
```diff
@@ -71,7 +69,7 @@ maths = []
 maths-nopanic = ["maths"]
 ndarray = ["dep:ndarray"]
 proptest = ["dep:proptest"]
-rand = ["dep:rand"]
+rand-0_9 = ["dep:rand-0_9"]
```
Tony-Samuels, line 72: My proposal is that for every 0.X optional dependency we have, they're specified & feature-flagged this way. That way as new versions of them come out, it's less confusing.

This also then reserves `rand` to be the 1.0 release.

Tony-Samuels: If this gets merged as is, I'll raise an MR to move the rest of them

paupino (approved):
Makes sense to me. Ultimately I think we need to provide guidance for dependencies/features going forward as part of the contributing guide. 

Side note; I'm not sure if this is the right breakdown, but perhaps there are three categories of dependencies:

1. Changing decimal behavior (e.g. math components, decimal only code)
2. Adding decimal behavior (e.g. rand)
3. Bridging decimal for use with another library (e.g. serialization etc)

The third one I think needs a bit of a play around - I'd love to investigate transparent types etc here, possibly in separate crates (or sub-crates). It may not be realistic, but ideally want to see if there is a way to provide a little bit more flexibility around how to approach the bridge (e.g. serde is a great example here).

But back to this PR - this looks great! If you do convert the other libraries to this format, perhaps we can start a doc to provide guidance for adding / maintaining / deprecating (inside `CONTRIBUTING.md`?). It doesn't need to be complete of course, just something we can start iterating on.

PR #726, opened 2025-04-30
As F64/AS i128

rollo-b2c2:


paupino:
I think what you're trying to achieve is right - I think for this PR you just need to take the arguments outside of the `ToPrimitive` implementation and treat them as a direct `impl` function.

PR #730, opened 2025-05-29
`postgres` -> `postgres_backend`

aumetra:
Analogue to #707 but for the `postgres` feature in Diesel.  
Makes stuff easier when exclusively using `diesel_async`.


Tony-Samuels:
If you raise a similar MR against `v1` I'll happily approve it and it will be made available in the next release.

PR #731, opened 2025-06-10
Fix panic on 0.x parsing

### LINKED ISSUE #715: Panic when placeholder is at rounding position and the integer part is zero
Panics when parsing: `0.1234567890123456789012345678_9`

The tests only check cases where the number start with `1.xxxxx` and they pass.

It only panics when the number start with `0.xxxxx` with `_` at the rounding position followed by number(s).

So I suggest adding tests starting with `0.xxxxx`.

Tony-Samuels:
This is not an area of code I've looked at previously, so this may not be a good fix. As per the added test case, it does resolve the reported issue.

Attempts to resolve #715.

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -357,8 +357,20 @@ fn handle_full_128<const POINT: bool, const NEG: bool, const ROUND: bool>(
                         if ROUND {
                             // If it is an underscore at the rounding position we require slightly different handling to look ahead another digit
                             if next == b'_' {
-                                if let Some((next, bytes)) = bytes.split_first() {
-                                    handle_full_128::<POINT, NEG, ROUND>(data, bytes, scale, *next)
+                                // Skip consecutive underscores to find the next actual character
+                                let mut remaining_bytes = bytes;
+                                let mut next_char = None;
+                                while let Some((n, rest)) = remaining_bytes.split_first() {
+                                    if *n != b'_' {
+                                        next_char = Some(*n);
+                                        break;
+                                    }
+                                    remaining_bytes = rest;
+                                }
+
+                                if let Some(ch) = next_char {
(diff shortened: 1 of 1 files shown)


src/str.rs
```diff
@@ -1035,6 +1039,14 @@ mod test {
                     28,
                 )),
             ),
+            (
+                "1.234567890123456789012345678_9",
+                Ok(Decimal::from_i128_with_scale(12345678901234567890123456789, 28)),
+            ),
+            (
+                "0.1234567890123456789012345678_9",
+                Ok(Decimal::from_i128_with_scale(1234567890123456789012345678, 28)),
```
paupino, line 1048: For this test, should we not expect `1234567890123456789012345679` (i.e. rounded up?).

paupino (commented):
Bear with me as I get up to speed with this! 

I notice in the test that:
```
0.12345678901234567890123456789
```
Turns into:
```
0.1234567890123456789012345678
```
Whereas, I think perhaps it should be (rounded up):
```
0.1234567890123456789012345679
```

I haven't looked into this much so don't have a good suggestion forward (yet) but I think this is ever so subtly incorrect right now. 

That said... a subtly incorrect value may be better than a panic.

Let me think a little bit more - the panic that is happening is from an assert but in essence having a scale of 29 may sometimes be ok.

paupino:
I've come up with an alternative solution which rounds, however want to ensure you retain credit so see if I can update your branch (and let you bring over to master if need be)

paupino:
Thank you @Tony-Samuels for taking the charge on getting this fixed! 🎉  

I'll merge and release a dot release for v1. If you can take the charge for getting this into v2 that'd be awesome 😄

PR #735, opened 2025-07-04
Add new align16 feature to force 16 byte alignment for Decimal

### LINKED ISSUE #649: Bug: not 128bit aligned with `repr-c` feature
I've noticed that while [docs](https://docs.rs/rust_decimal/1.31.0/rust_decimal/index.html#c-repr) claim that `Decimal` is 128bit aligned it's not the case which could lead to UB ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e32abe7612da2f558fed71ab7cc19288) run with miri).

Either docs should be edited of fix implemented. 

I can work on the fix by introducing new ZST field `_align: [u128; 0]` into `Decimal` which would force proper alignment of the type to be equal to `u128`.

```rust
#[cfg_attr(feature = "c-repr", repr(C))]
pub struct Decimal {
    // Bits 0-15: unused
    // Bits 16-23: Contains "e", a value between 0-28 that indicates the scale
    // Bits 24-30: unused
    // Bit 31: the sign of the Decimal value, 0 meaning positive and 1 meaning negative.
    flags: u32,
    // The lo, mid, hi, and flags fields contain the representation of the
    // Decimal value as a 96-bit integer.
    hi: u32,
    lo: u32,
    mid: u32,
    // Enforce proper alignment
    #[cfg(feature = "c-repr")]
    _align: [u128; 0],
}
```

Tony-Samuels:
Fixes #649, now that the underlying alignment of 128 bit numbers has been fixed by rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ tokio-postgres = { default-features = false, version = "0.7" }
 [features]
 default = ["serde", "std"]
 
+align16 = [] # Force Decimal to be repr(align16) - the same as u128
 borsh = ["dep:borsh", "std"]
 c-repr = [] # Force Decimal to be repr(C)
 db-diesel-mysql = ["diesel/mysql_backend", "std"]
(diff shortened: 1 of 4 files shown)


Tony-Samuels:
Once this is merged, I'll push something to master too.

paupino (approved):
LGTM - Thank you!

PR #739, opened 2025-07-27
Improved exp

Danvil:
Improved algorithm to compute exp (and thus also pow) to not compute
x^i directly in the Taylor series as it quickly outgrows exp(x).
Instead the summand is grown with x/i. This also avoid using the
FACTORIAL table. This enables computation up to `exp(66.5)` instead
of previously ~`exp(11.8)`.

Additionally the loop avoids two unnecessary ops for the tolerance
check. In total performance improved by about 5%.

As part of the new method the total iteration count of the algorithm
was increased to better meet the given tolerance. This only affects
computation for large x. Bench test shows about 4x less performance
for x > ~18.


Tony-Samuels:
Looks like a few of the test cases have slightly different results than your PR set them to.

Tony-Samuels:
I'll run a few benchmarks myself tomorrow and dig into the perf differences bit more, but if this is both a perf improvement at low scales and also allows for use with larger numbers then it seems like a pretty safe inclusion. 

Thank you.

PR #742, opened 2025-08-13
Add `#[inline]` to `Decimal::from_parts` for performance

finnbear:
This makes [decoding `Decimal` in `bitcode`](https://github.com/SoftbearStudios/bitcode/pull/42) 10X faster.

It works since `bitcode` checks the range of limited-range integers (like `scale`) up front, all at once, using vector instructions, and then uses `unreachable_unchecked` right before calling `Decimal::from_parts` to communicate that fact to `rustc`. This only works if `Decimal::from_parts` is inlinable.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -522,6 +522,7 @@ impl Decimal {
     /// let pi = Decimal::from_parts(1102470952, 185874565, 1703060790, false, 28);
     /// assert_eq!(pi.to_string(), "3.1415926535897932384626433832");
     /// ```
+    #[inline]
     #[must_use]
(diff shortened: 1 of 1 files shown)


finnbear:
> I'm surprised the compiler doesn't just do this TBH, the inline attribute is only a suggestion.

AFAIK, the Rust compiler doesn't automatically inline non-generic functions across crate boundaries unless using LTO (link-time optimization), which increases compile times and isn't on by default.

> If you'd like to see this in the next v1 release, please raise the same MR against the v1 branch. 

Good to know, I'll keep that in mind!

PR #746, opened 2025-09-04
Add from_scientific_round, that parses like from_scientific but behaves like from_str for numbers with greater precision that Decimal supports

tenuous-guidance:


src/decimal.rs
```diff
@@ -637,6 +637,102 @@ impl Decimal {
...
+    ///
+    /// let value = Decimal::from_scientific_round("-2.5e-28")?;
+    /// assert_eq!(value.to_string(), "-0.0000000000000000000000000003");
+    ///
+    /// let err = Decimal::from_scientific_round("2e-29").unwrap_err();
+    /// assert_eq!(err, Error::ScaleExceedsMaximumPrecision(29));
+    /// #     Ok(())
+    /// # }
+    /// ```
+    pub fn from_scientific_round(value: &str) -> Result<Decimal, Error> {
```
paupino, line 670: I'm not a big fan of this name, but I get where you're coming from. What about `from_scientific_lossy`? I only pick that up since it loosely matches a Rust naming convention (e.g. `from_utf8_lossy`).

paupino (commented):
Overall, this looks good and makes sense. I added one comment - am certainly open to suggestions.

paupino:
Yes, good idea! It's been on my to do list for a little while -- I'll see if I can get something published this weekend.

PR #750, opened 2025-09-17
Allow the usage of fallible operations in constant environments

c410-f3r:
Finishes what was started at #749

Feel free to indicate things that should be modified.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -394,7 +394,7 @@ impl Decimal {
     #[must_use]
     pub fn new(num: i64, scale: u32) -> Decimal {
         match Self::try_new(num, scale) {
-            Err(e) => panic!("{}", e),
+            Err(e) => panic!("{e}"),
             Ok(d) => d,
         }
     }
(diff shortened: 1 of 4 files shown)


macros/src/str.rs
```diff
@@ -76,8 +76,8 @@ impl Display for ParseError<'_> {
     }
 }
 
-pub const fn parse_decimal(src: &str, exp: i32) -> ParseResult {
-    const fn skip_us(radix: u32, is_positive: bool, mut src: &[u8], exp: i32) -> ParseResult {
+pub const fn parse_decimal(src: &str, exp: i32) -> ParseResult<'_> {
+    const fn skip_us(radix: u32, is_positive: bool, mut src: &[u8], exp: i32) -> ParseResult<'_> {
```
c410-f3r, line 80: Fixes a warning

```
lifetime flowing from input to output with different syntax can be confusing
`#[warn(mismatched_lifetime_syntaxes)]` on by default
```

Issue #722, opened 2025-04-20
from_str should support scientific notation


blastrock:
Hi,

I have a case where numbers might be in classical notation, or in scientific notation for very small numbers. I understand that `from_str` handles the first case, and `from_scientific` the second. This forces me to write code like:

```rust
Decimal::from_str(fields[1])
                            .or_else(|_| Decimal::from_scientific(fields[1]))
```

It would help me if from_str just handles any notation.

Tony-Samuels:
I suspect this could fit nicely into the 2.0 branch.

PR #753, opened 2025-10-04
Fixed issue #722 to support scientific notation in Decimal::from_str()

kofki:
Solving issue #722 

Implemented notation handling in the `from_str` function of `Decimal` to be able to parse through scientific or decimal notation.  It calls `from_scientific` function within `from_str`, handling both cases of notation.

Testing:
> Passed all the testsScreenshot 2025-10-04 001101

> Added two test for scientific notation parsing handling Screenshot 2025-10-04 001110

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1981,7 +1981,17 @@ impl FromStr for Decimal {
     type Err = Error;
 
     fn from_str(value: &str) -> Result<Decimal, Self::Err> {
-        crate::str::parse_str_radix_10(value)
+        match crate::str::parse_str_radix_10(value){
+		Ok(d) => Ok(d),
+
+		Err(e) => {
+			if value.contains('e') || value.contains("E") {
+				Decimal::from_scientific(value)
+			} else{
+				Err(e)
+			}
+		}
+	}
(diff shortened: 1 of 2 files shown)


src/decimal.rs
```diff
@@ -1981,7 +1981,17 @@ impl FromStr for Decimal {
     type Err = Error;
 
     fn from_str(value: &str) -> Result<Decimal, Self::Err> {
-        crate::str::parse_str_radix_10(value)
+        match crate::str::parse_str_radix_10(value){
+		Ok(d) => Ok(d),
+
+		Err(e) => {
+			if value.contains('e') || value.contains("E") {
```
Tony-Samuels, line 1988: Given that `e` and `E` are not valid base10 digits, I suspect we could use this to determine which parsing function to use immediately, without having to try `parse_str_radix_10` first.

paupino:
Sorry, I inadvertently closed this - effectively I found a significant slowdown with checking first and meant to revert the change but instead closed this PR. We need to be cautious here since we want to optimize for the most common case first (i.e. decimal strings vs scientific stings).

PR #759, opened 2025-10-22
Support `no-std::no-alloc`

onkoe:
This isn't tested since my Mac has low storage. But here's a little peek into how `no-alloc` support could be done.

Thanks for the great crate! :)

## Context

I want to use the `rust_decimal` crate in my own library, but the lack of `core-only` support is a blocker to something I want to do.

## Checklist

- [x] Add test crate with the following setup:
	- Depend on `rust_decimal` in Cargo.toml
	- Have a veeery basic `lib.rs` and `pub fn t() -> Decimal { /* ... */ }`
	- Build with: `cargo +nightly build --no-default-features -Z build-std=core --target x86_64-unknown-none`
		- Need to pin the Nightly
- [x] Document `no_std` + `core`-only compatibility in `README.md`
- [x] Make block comments in `error.rs` look consistent (and maybe even readable)
- [x] Fix "system dependency required" tests expecting stringy errors

## Unrelated

A couple more suggestions before 2.0:

- Move all the integrations into `src/integrations/(...).rs` files
	- You can re-export them in `lib.rs` if you'd like to maintain the same API surface
		- `mod integrations; pub use integrations::*;`
	- But, as-is, it's difficult to focus on the primary logic files in the crate
- Rename `Error` to `DecimalError` 
	- Per Clippy: https://rust-lang.github.io/rust-clippy/master/index.html#error_impl_error

### DIFF
--- a/make/tests/misc.toml
+++ b/make/tests/misc.toml
@@ -1,11 +1,16 @@
 [tasks.test-misc]
 dependencies = [
+    "test-no-alloc",
     "test-proptest",
     "test-rust-fuzz",
     "test-rocket-traits",
     "test-borsh",
     "test-rand"
 ]
+
+[tasks.test-no-alloc]
+command = "cargo"
+args = ["build", "--no-default-features"]
 
 [tasks.test-proptest]
 command = "cargo"
--- a/make/tests/misc.toml
+++ b/make/tests/misc.toml
@@ -9,8 +9,13 @@ dependencies = [
 ]
 
 [tasks.test-no-alloc]
+dependencies = ["install-no-alloc-target"]
 command = "cargo"
-args = ["build", "--no-default-features"]
+args = ["build", "--no-default-features", "--target", "x86_64-unknown-none"]
+
+[tasks.install-no-alloc-target]
+command = "rustup"
+args = ["target", "add", "x86_64-unknown-none"]
 
 [tasks.test-proptest]
 command = "cargo"
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -567,24 +567,20 @@ impl Decimal {
     pub fn from_scientific_exact(value: &str) -> crate::Result<Decimal> {
         let mut split = value.splitn(2, ['e', 'E']);
 
-        let base = split.next().ok_or(crate::Error::FailedToParseScientificFromString)?;
-        let exp = split.next().ok_or(crate::Error::FailedToParseScientificFromString)?;
+        let base = split.next().ok_or(crate::Error::ScientificBaseNotFound)?;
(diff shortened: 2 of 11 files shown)


src/decimal.rs
```diff
@@ -2453,12 +2450,14 @@ impl fmt::Debug for Decimal {
     }
 }
 
+#[cfg(any(feature = "alloc", feature = "std"))]
```
Tony-Samuels, line 2453: Nit: If std requires alloc, then we only need to test for alloc.

Tony-Samuels (approved):

make/tests/misc.toml
```diff
@@ -1,12 +1,17 @@
 [tasks.test-misc]
 dependencies = [
```
Tony-Samuels, line 2: Might be worth considering pulling cargo hack into our test infra. cargo hack check --feature-powerset has been quite useful in other contexts to detect feature combination breakages.

make/tests/misc.toml
```diff
@@ -1,12 +1,17 @@
...
     "test-rust-fuzz",
     "test-rocket-traits",
     "test-borsh",
     "test-rand"
 ]
 
+[tasks.test-no-alloc]
+command = "cargo"
+args = ["build", "--no-default-features"]
+
```
onkoe, line 14: Please double-check that this builds in a `no_std` environment. With this command, `rustc` is still linking in `libstd` for the host target (i.e., the one building the crate). This means that this test build tests that the features build -- not that they actually work on `no_std` targets!

I recommend either:

- building for a T3 target that doesn't support `std` and `alloc`, or
- compiling the STL from scratch (`build-std=core`) with `libcore` only via a pinned Nightly version.

The first option is likely easier for this project. You can quickly and easily install it via `rustup` in the environment. If you want this in another PR, please let me know and I'm happy to do just that :D

onkoe: The [`x86_64-unknown-none`](https://doc.rust-lang.org/nightly/rustc/platform-support/x86_64-unknown-none.html) target should work for option 1! Please let me know if you need any more info

src/decimal.rs
paupino: Fixed

make/tests/misc.toml
paupino: That seems like a useful feature - let's create a new PR, perhaps targeting V1/master to add this (btw, assume it's this: https://github.com/taiki-e/cargo-hack?)

make/tests/misc.toml
paupino: Added - of course we can continue iterating on this too as required.

PR #772, opened 2026-01-14
Replace ryu with zmij for float formatting

hugobast:
## Summary
- Replaces `ryu` with `zmij` for floating-point to string conversion in serde arbitrary precision deserialization
- Follows the same direction as serde_json: https://github.com/serde-rs/json/pull/1304

## Related
- https://github.com/paupino/rust-decimal/pull/761


Tony-Samuels:
https://github.com/serde-rs/json/commit/1aefe152735f1b11ce7f641f8e86448d227163bf
serde-json updated when 1.0 was released, although it was done via a direct commit to master as far as I can tell, and not a PR.

hugobast:
> serde-json updated when 1.0 was released

Great! Thanks for flagging.

Issue #431, opened 2021-10-08
`powd` overflow


maurolacy:
Hello, and first off, thanks for the library.

We're using `rust_decimal` to compute a function using fixed point math, involving `exp` and `powd`.

Now, what we are noticing is that `powd`'s range is not very large. By example, when raising a `Decimal` larger than 32_313_447 to a Decimal power smaller than one (0.68 in particular) using `powd`, we are hitting

`thread ... panicked at 'Pow overflowed', .../github.com-1ecc6299db9ec823/rust_decimal-1.16.0/src/maths.rs:283:21`

This doesn't make much sense, as a number raised to an exponent smaller than 1 would yield a _smaller_ number (127_755, in this particular case).

Are you aware of this? I can always take a look at the code, but, asking first in case this is a kwown issue / there are known workarounds. What do you suggest? 

Thanks,

maurolacy:
Also, I've noticed that `sqrt()` doesn't suffer from this. Or at least, its range of application is order of magnitudes larger than `powd`. So, `powd` can for sure be made to behave, at least for exponents smaller than one.

schungx:
> We cannot. We are forbidden to use floating point at all, because of non-determinism, i.e. different results being obtained
> in different implementations / architectures.

I understand that the IEEE 754 standard defines deterministic calculations for addition, subtraction, multiplication, division plus the square root, meaning that they should generate deterministic output when given the exact same inputs.

For exponential (and also power), you start running into the Table Maker's Dilemma and sooner or later you get into rounding situations.

It is probably possible to make an `exp` or `pow` where, for each round, it then normalizes the number into a fixed number of significant digits before continuing... which may make it accept a wider range of inputs at the risk of having rounding errors starts accumulating.

I'm not really an expert on this, but if you want `exp` or `pow` to work over a full range of 32-bit numbers, you may need more than the 96 bits that this crate is keeping.  In fact, you may need to use a arbitrary-length number library.

schungx:
https://github.com/paupino/rust-decimal/blob/master/src/maths.rs#L316-323

This is where it calculates the `pow` by converting it into `exp`, and this may be the place where it overflows.

Since you know the exponent will be <1, I suggest you break down the steps and display intermediate values to find out which step it is overflowing.  Then you may simply add `normalize` calls?

maurolacy:
[...]

Issue #690, opened 2024-11-08
powd / powf result differ from regular rust pow and other online calculators


SamuelRamond:
I was wondering if this miss match vs rust regular pow in result was expected ?

```rust
println!("{}", dec!(2.).powd(dec!(3.2)));
# 9.189586821131990468278294299
println!("{}", dec!(2.).powf(3.2));
# 9.189586821131990468278294299

let a: f64 = 2.;
println!("{}", a.powf(3.2));
# 9.18958683997628
```

also checking a few calculator online seems to be in favor of regular rust response:
<img width="813" alt="image" src="https://github.com/user-attachments/assets/3f189f0f-a61c-44bc-a5df-6ca86525af9a">
<img width="894" alt="image" src="https://github.com/user-attachments/assets/f598ff12-1e9d-4b69-8e1a-40c1297c9d27">
<img width="641" alt="image" src="https://github.com/user-attachments/assets/2c3cac8f-cfc3-4d63-80b2-e0f68e75d457">
Iphone calculator also match google and co

Is there something i'm missing ?

AndreaCensi:
I also encountered this issue when porting code from Python (`decimal` crate) to Rust.

Consider the operation:

`908683.5 ^ (2/3)` 

In Python I get:

```
908683.5 ^ (2/3) = 9381.560741153 
```
which is a good answer because
```
9381.560741153 ^ (3/2) =  908683.5 
```

Google gives the same answer with fewer significant digits.


With this crate I get 

```
908683.5 ^ (2/3) =  9381.556826468 
```

which we can see it is off because if we invert it, we get

```
 9381.556826468  ^ (3/2) = 908682.931244
```



I tried to look into how this operation is implemented in Python. The `decimal` module is implemented in C. These are the sources: https://github.com/python/cpython/tree/main/Modules/_decimal/libmpdec
The code is quite opaque and I couldn't find it. Maybe somebody else can find it.

AndreaCensi:
Digging a bit deeper, I found out that most of the pow logic is written in Python here:

https://github.com/python/cpython/blob/27c0d9b54abaa4112d5a317b8aa78b39ad60a808/Lib/_pydecimal.py#L2432C9-L2451C65

The code checks for a multitude of special cases, then it uses the same formula as this crate - x**y = exp(y*log(x))

```
 // We do some approximations since we've got a decimal exponent.
        // For positive bases: a^b = exp(b*ln(a))
        let negative = self.is_sign_negative();
[...]

Issue #471, opened 2022-01-28
Decimal precision mismatches


apalepu23:
Hey all,

I know there have been some conversations around precision (re: #414 ), but wanted to post a different question on the topic for guidance/support:

CONTEXT:
We have a Rust codebase that performs computations (using this `rust-decimal` module) and outputs a record of the inputs to these computations. We have a Python sister codebase that receives this log and attempts to perform the exact same computations (using the [Decimal](https://docs.python.org/3/library/decimal.html#module-decimal) module) to arrive at the same results. 

We noticed that on occasion, there are ever so slight discrepancies.

REPRO:

FYI: For Python, we set the `Decimal` module's context to be the following:

```
getcontext().prec = 29
getcontext().Emin = 0
getcontext().Emax = 28
getcontext().clamp = 1
```

We chose these values after studying the `Decimal` module and attempting to match how `rust-decimal` handles arithmetic. In particular, we believe these `Emin` and `Emax` values give us an `e` in the range of [-28, 0], and the `prec` setting of 29 (as opposed to the default value of 28) would allow us to represent 29 digits since `rust-decimal` can support numbers as large as 2 ** 96 - 1 (which has 29 digits).

**Case 1 (Matches) ==> 5631989.747461568422879160 + 12354.867325583148587639999525:**
```
// RUST
Decimal::from_str("5631989.747461568422879160").unwrap() + Decimal::from_str("12354.867325583148587639999525").unwrap() == 5644344.6147871515714667999995

// PYTHON
Decimal("5631989.747461568422879160") + Decimal("12354.867325583148587639999525") == 5644344.6147871515714667999995
```

**Case 2 (Matches) ==> 6631989.747461568422879160 + 12354.867325583148587639999525:**
```
// RUST
Decimal::from_str("6631989.747461568422879160").unwrap() + Decimal::from_str("12354.867325583148587639999525").unwrap() == 6644344.6147871515714667999995

// PYTHON
Decimal("6631989.747461568422879160") + Decimal("12354.867325583148587639999525") == 6644344.6147871515714667999995
```

**Case 3 (Matches) ==> 7631989.747461568422879160 + 12354.867325583148587639999525:**
```
// RUST
Decimal::from_str("7631989.747461568422879160").unwrap() + Decimal::from_str("12354.867325583148587639999525").unwrap() == 7644344.6147871515714667999995

// PYTHON
[...]

PR #776, opened 2026-03-04
Widen mathematical operations to retain precision as long as possible during calculations

paupino:
Previously, during mathematical operations which leverage a loop we would effectively widen the calculation and then narrow it again each iteration of the loop (i.e. widen for multiplication to 192 bits, and then narrow it back to 96 bits every iteration). This was something that always bothered me since we're effectively losing precision every iteration.

This PR replaces the 96-bit intermediate arithmetic in all `maths` functions (`exp`, `ln`, `sin`, `cos`, `powu`) with 192-bit wide arithmetic (`DecWide`). This doubles the working precision during iterative calculations, only truncating to 96-bit `Decimal` at the final result. Because we have higher precision (i.e. don't need to converge as early) also performed som algorithm improvements (argument reduction for `exp`, `atanh` series for `ln`). This fixes long-standing precision bugs where results were wrong from digit 9-13, and now match Python's 50-digit Decimal to 25-28 digits.

Performance wise, this is an overall improvement while still fixing a few outstanding issues.

```
| Function | Old (96-bit, ns)| New (192-bit, ns)| Change               |
|==========|=================|==================|======================|
| powi     | 268             | 258              | -4%                  |
| erf      | 5,953           | 2,928            | -51%                 |
| exp      | 9,243           | 7,162            | -23%                 |
| exp_large| 10,656          | 9,113            | -14%                 |
| ln       | panicked        | 18,092           | fixed                |
| norm_cdf | 15,238          | 7,154            | -53%                 |
| norm_pdf | 10,121          | 7,455            | -26%                 |
| sqrt     | 4,431           | 4,431            | same                 |
```

Fixes: #431, #471, #690


Makefile.toml
```diff
@@ -8,3 +8,36 @@ extend = [
 
 [config]
 default_to_workspace = false
+
+# Runs the same checks as GitHub Actions CI (excluding database and fuzz tests
+# which require external services or nightly toolchain + cargo-fuzz).
+# Usage: cargo make ci
+[tasks.ci]
```
paupino: I'm not sure if this is useful or not. I just wanted a quick way to check for common issues before pushing up

tests/decimal_tests.rs
```diff
@@ -4141,27 +4103,27 @@ mod maths {
         let test_cases = &[
             (
                 Decimal::from_str("-2.0").unwrap(),
-                Decimal::from_str("0.0539909667221995238993056051").unwrap(),
+                Decimal::from_str("0.0539909665131880519505642004").unwrap(),
```
paupino: These now align with python decimal

PR #785, opened 2026-03-23
fix: scientific formatting of 0

### LINKED ISSUE #784: `format!("{:e}", Decimal::ZERO)` produces invalid scientific notation
The following program demonstrates the problem:

```rust
use rust_decimal::Decimal;

fn main() {
    let d = Decimal::ZERO;
    let sci = format!("{:e}", d);
    println!("format!(\"{{:e}}\", Decimal::ZERO) = {:?}", sci);
    let parsed = Decimal::from_scientific(&sci);
    println!("Decimal::from_scientific({:?}) = {:?}", sci, parsed); 
}
```

This prints:

```
format!("{:e}", Decimal::ZERO) = "e0"
Decimal::from_scientific("e0") = Err(ErrorString("Invalid decimal: empty"))
```

The "e0" produced isn't valid scientific notation (I think the correct thing here would just be "0", though I guess "0e0" would also be valid, but you need *something* before that e).

Tony-Samuels:
I've just added an `::is_zero` short-circuit, which I think makes sense here. Alternatively, we could do extra checks in a few places, but I this is probably simpler. Otherwise there will need to be a few checks to avoid results like `0e6`.

Resolves #784

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2795,3 +2795,14 @@ impl<'a> Sum<&'a Decimal> for Decimal {
         sum
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_scientific_0e0() {
+        let dec = Decimal::from_scientific("0e0").unwrap();
+        assert_eq!(dec, Decimal::ZERO);
+    }
+}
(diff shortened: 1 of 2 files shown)


paupino (approved):

paupino:
Thank you!

PR #794, opened 2026-04-25
Optimize core arithmetic operations (+, -, *, /)

paupino:
Took a pass at the core arithmetic hot paths and found some solid wins; ~56% reduction in total benchmark time across `add`, `sub`, `mul`, `div`, and `rem`. All internal changes.

 ### Benchmark results

| Operation | Before (ns/iter) | After (ns/iter) | Change |
|-----------|-----------------|-----------------|--------|
| `add_self` | 3.22 | 1.26 | -61% |
| `add_10k` (fold) | 88,695 | 45,085 | -49% |
| `sub_10k` (fold) | 88,586 | 45,311 | -49% |
| `mul_25` (fold) | 229 | 141 | -38% |
| `div_10k` (fold) | 35,946 | 7,497 | -79% |
| `rem_10k` (fold) | 34,883 | 11,380 | -67% |
 
Individual ops like `mul_one` and `div_pi` are roughly unchanged; the wins are concentrated in the paths that were paying unnecessary overhead per call.

### What changed (and why)

**Inline hot arithmetic functions.** This was the single biggest win (~31% on its own). `add_impl`, `sub_impl`, `mul_impl` and their internal helpers (`fast_add`, `aligned_add`, `Dec64::new`, `Dec64::to_decimal`) were not being inlined across module boundaries. Adding `#[inline(always)]` eliminates call overhead on every operation. `div_impl` gets `#[inline]` rather than `#[inline(always)]` since it's large enough that forcing it everywhere hurts icache.

**Skip redundant checks in result construction.** Arithmetic operations already validate inputs at entry (zero checks, etc.), so the result constructor doesn't need to re-verify. `from_parts_raw_unchecked` skips the zero-sign normalization when we know the result is non-zero; `Dec64::to_decimal` uses it along with a `pub(crate) flags()` helper to avoid the scale assertion in `from_parts`.

**Branchless Dec64::new.** The constructor previously branched on `mid == 0` to skip a shift. Removing the branch and always computing `(mid as u64) << 32 | lo as u64` turns out to be faster; the shift-with-zero is essentially free, but the branch was causing mispredictions in div/rem paths that construct Dec64 values with varying mantissa sizes.

**Restructure `add_sub_internal`.** The 32-bit fast path (both operands fit in u32) is checked before the full `is_zero()` test. When `mid` and `hi` are already known to be zero, the zero check reduces to a single `lo == 0` comparison instead of loading and OR-ing three fields. This shaves a few cycles off the most common path in fold operations.

**Lazy Buf24 initialization in `mul_impl`.** The 24-byte `Buf24` buffer was being zeroed on every multiply, even when the common 32x32 fast path returns before touching it. Moving the allocation into the branches that actually need it (32x64, 32x96, 96x96) avoids unnecessary stack setup. This also required extracting a `finish_mul` helper since product is scoped per-branch.

### Things I tried that didn't help

Worth noting for anyone revisiting this later:

- Additional `#[inline(always)]` on `is_zero`, `scale`, `from_parts` had no effect; the compiler already inlines `const fn` within the same crate
- Branchless `from_parts_raw` using a `wrapping_neg` mask was slower; the branch is well-predicted and cheaper than the mask computation
- An "ultra-fast" dedicated path for same-flags + both-32-bit duplicated too much code and hurt icache
- Extracting the slow path (>32-bit operands) into `#[inline(never)]` made things worse; that path is used frequently enough (64-bit aligned adds, fold operations) that the function call overhead dominates


paupino:
Complete set of tests:


Test | Before (ns/iter) | After (ns/iter) | Change
-- | -- | -- | --
`add_10k` | 85,840.72 | 48,534.12 | -43.5%
`add_64bit_aligned` | 3.73 | 2.00 | -46.4%
`add_96bit_aligned` | 3.77 | 2.00 | -46.9%
`add_96bit_sub` | 3.52 | 2.25 | -36.1%
`add_large_scale_diff` | 21.23 | 21.58 | +1.6%
`add_negative_pi` | 6.94 | 6.00 | -13.5%
`add_negative_point_five` | 3.47 | 1.86 | -46.4%
`add_one` | 3.47 | 1.56 | -55.0%
`add_one_hundred` | 3.47 | 1.56 | -55.0%
`add_pi` | 6.83 | 5.90 | -13.6%
`add_point_zero_one` | 3.10 | 1.24 | -60.0%
`add_self` | 3.11 | 1.24 | -60.1%
`add_simple` | 3.10 | 1.24 | -60.0%
`add_two` | 3.48 | 1.58 | -54.6%
`decimal_from_str` | 71.40 | 72.58 | +1.7%
`decimal_to_string` | 646.04 | 641.36 | -0.7%
`div_10k` | 34,834.12 | 7,196.27 | -79.3%
`div_32bit_divisor` | 4.34 | 2.80 | -35.5%
`div_64bit_divisor` | 3.95 | 2.80 | -29.1%
`div_96bit_divisor` | 11.23 | 10.36 | -7.7%
`div_negative_pi` | 44.21 | 47.28 | +6.9%
`div_negative_point_five` | 19.13 | 15.65 | -18.2%
`div_no_underflow` | 36.02 | 36.69 | +1.9%
`div_one` | 4.36 | 2.80 | -35.8%
`div_one_hundred` | 19.31 | 15.80 | -18.2%
`div_pi` | 44.50 | 47.29 | +6.3%
`div_point_zero_one` | 4.02 | 2.80 | -30.3%
`div_repeating` | 21.14 | 21.02 | -0.6%
`div_self` | 5.18 | 4.10 | -20.8%
`div_two` | 19.16 | 14.41 | -24.8%
`iterator_individual` | 32.11 | 15.65 | -51.3%
`iterator_product` | 46.88 | 28.75 | -38.7%
`iterator_sum` | 33.94 | 14.76 | -56.5%
`mul_25` | 222.85 | 60.53 | -72.8%
`mul_64bit_by_32bit` | 2.81 | 2.52 | -10.3%
`mul_64bit_by_64bit` | 3.16 | 2.83 | -10.4%
`mul_96bit_by_32bit` | 8.75 | 7.06 | -19.3%
`mul_high_scale` | 2.99 | 2.51 | -16.1%
`mul_negative_pi` | 8.76 | 7.05 | -19.5%
`mul_negative_point_five` | 2.53 | 2.21 | -12.6%
`mul_one` | 2.53 | 2.23 | -11.9%
`mul_one_hundred` | 2.53 | 2.25 | -11.1%
`mul_overflow_rescale` | 41.56 | 40.79 | -1.9%
`mul_pi` | 8.77 | 7.12 | -18.8%
`mul_point_zero_one` | 2.52 | 2.19 | -13.1%
`mul_two` | 2.52 | 2.21 | -12.3%
`rem_10k` | 33,603.23 | 11,692.28 | -65.2%
`serialize_bincode` | 785.85 | 786.62 | +0.1%
`sub_10k` | 85,857.95 | 45,627.65 | -46.9%
`sub_64bit_aligned` | 3.75 | 2.27 | -39.5%
`sub_large_scale_diff` | 23.20 | 23.27 | +0.3%
`sub_negative_pi` | 7.15 | 5.92 | -17.2%
`sub_negative_point_five` | 3.48 | 1.58 | -54.6%
`sub_one` | 3.47 | 1.57 | -54.8%
`sub_one_hundred` | 3.48 | 1.75 | -49.7%
`sub_pi` | 7.02 | 6.20 | -11.7%
`sub_point_zero_one` | 3.42 | 1.55 | -54.7%
`sub_self` | 3.43 | 1.58 | -53.9%
`sub_simple` | 3.44 | 1.56 | -54.7%
`sub_two` | 3.49 | 1.58 | -54.7%

There are some regressions in a couple of div tests (e.g. `div_pi`, `div_negative_pi`), however likely within the noise margins.

src/ops/mul.rs
```diff
@@ -124,8 +129,14 @@ pub(crate) fn mul_impl(d1: &Decimal, d2: &Decimal) -> CalculationResult {
         } else {
             product.set_mid64(tmp);
         }
+        return finish_mul(product, negative, scale);
     }
+    // All branches above return, so this is unreachable
+    unreachable!()
```
paupino: Triggers a warning - perhaps remove?

src/decimal.rs
```diff
@@ -549,6 +549,13 @@ impl Decimal {
     }
 
     #[must_use]
+    /// Constructs a Decimal without any zero-sign normalization.
+    /// Caller must guarantee that if lo|mid|hi == 0, the sign bit in flags is cleared.
+    #[inline(always)]
+    pub(crate) const fn from_parts_raw_unchecked(lo: u32, mid: u32, hi: u32, flags: u32) -> Decimal {
```
paupino, line 555: There is an argument to apply this to `from_parts_raw` too. It was to avoid `-0` being a thing, which we could perhaps address differently.

src/decimal.rs
```diff
@@ -1898,8 +1905,8 @@ pub(crate) enum CalculationResult {
     DivByZero,
 }
 
-#[inline]
-const fn flags(neg: bool, scale: u32) -> u32 {
+#[inline(always)]
+pub(crate) const fn flags(neg: bool, scale: u32) -> u32 {
```
paupino, line 1907: I think this is probably getting inlined anyway, however we should expand the visibility and assert that it should always be inlined.

src/ops/add.rs
```diff
@@ -2,65 +2,76 @@ use crate::constants::{MAX_I32_SCALE, POWERS_10, SCALE_MASK, SCALE_SHIFT, SIGN_M
...
+        let mut result = *d2;
+        if subtract && !d2.is_zero() {
+            result.set_sign_negative(d2.is_sign_positive());
+        }
+        return CalculationResult::Ok(result);
+    }
+    if d2.is_zero() {
+        return CalculationResult::Ok(*d1);
+    }
+
```
Tony-Samuels, line 76: Is it that much slower to do the zero check on >32-bits? If not, could combine the checks.

Tony-Samuels: In fact, we have to calculate whether the upper and mid parts are 0 anyway, to know that we can do 32-bit maths. So we could store those results (rather than using them immediately) and just do the 32-bit 0 check?

src/ops/mul.rs
Tony-Samuels: I think so? Unless rustc is forcing you to return a result, which it shouldn't be, then leaving it empty should make it just as clear than we've already return our result.

src/decimal.rs
Tony-Samuels: You probably know but just in case, `inline(always)` docs state: "This strongly suggests that the function should be inlined". There's no way to force inlining, the compiler will always decide.

Tony-Samuels (approved):
Looks fine to me, I'm surprised some of these make much of a difference, but I don't see any harm from them.

src/decimal.rs
paupino: Yeah, for sure - just wanting to _strongly_ hint 😂

src/ops/mul.rs
paupino: Will remove it to keep things clean.

src/decimal.rs
paupino: Ignoring for now, it is unchecked and consequently needs to be opted in.

src/ops/add.rs
paupino: I think it's a good callout. Previously it was optimizing for the zero path (adding zero, or subtracting zero) which does have a noticeable difference on 32 bit operations. 

I think further restricting the comparisons here is an interesting challenge though. Let me explore if it makes a feasible difference.

paupino: It's an interesting set of results. The added cache definitely has overhead, perhaps not worth it:

| Test               | Before    | After     | Change |
|--------------------|-----------|-----------|--------|
| `add_point_zero_one` | 1.24 ns   | 1.56 ns   | +25.8% |
| `add_self`           | 1.24 ns   | 1.57 ns   | +26.6% |
| `add_simple`         | 1.24 ns   | 1.55 ns   | +25.0% |
| `add_64bit_aligned`  | 1.98 ns   | 1.87 ns   | -5.6%  |
| `add_96bit_aligned`  | 1.97 ns   | 1.87 ns   | -5.1%  |
| `add_10k`            | 48,583 ns | 49,312 ns | +1.5%  |

Tony-Samuels: Huh. Very much not what I expected, but my intuition for micro-optimisations isn't that finely honed.

Tony-Samuels: I might play around with it a bit more at some point, but no reason not to merge the improvements we have,

paupino: I've often found that the things that I don't think will make much difference do, and the things that I think will make big differences don't 😂

PR #800, opened 2026-05-11
feat: implement num_trait gaps Bounded, NumCast, ConstZero, ConstOne…

### LINKED ISSUE #573: Add support for use num_traits::cast::NumCast
Disclaimer I am pretty new to rust

I was wondering if support could be added for the num_traits::cast::NumCast

The trait is defined as part of this file: https://docs.rs/num-traits/latest/src/num_traits/cast.rs.html#721

I would like to implement it as I wish to use Decimal with the geo library:  geo::Coord::<Decimal>::{x: 1., y:1.}
however it only accepts types which implement numcast.

 
I had a go at implementing how I think it should look , I think there are three high level options:

1) Figure out the input type and call the appropriate conversion method (requires figuring out input type)
2) Always convert input to an f64 and convert f64 to a Decimal (has some data loss when input is i128)
3) Use some magic I am unaware of 

I was wondering if I could get some pointers, or help implementing this.

In regards to figuring out the type, It looks like the standard approach here will not work, since we can convert the types to any of the primitive numbers, so checking if we can convert isn't fantastic. https://doc.rust-lang.org/std/any/index.html

I found a hacky looking, but powerful(?) approach [here ](https://www.reddit.com/r/rust/comments/mjfdif/is_it_possible_to_statically_check_generic_type/)

which I modified and included below, but the method I use to check types is maybe dangerous:

`
use num_traits::cast::{FromPrimitive, NumCast, ToPrimitive};
use crate::prelude::Decimal;
use std::any::type_name;

// By directly calling `type_name` here we guarantee that the names
// remain "up to date".
const I8_NAME: &str = type_name::<i8>();
const I16_NAME: &str = type_name::<i16>();
const I32_NAME: &str = type_name::<i32>();
const I64_NAME: &str = type_name::<i64>();
const I128_NAME: &str = type_name::<i128>();

const ISIZE_NAME: &str = type_name::<isize>();

const F32_NAME: &str = type_name::<f32>();
const F64_NAME: &str = type_name::<f64>();

const U8_NAME: &str = type_name::<u8>();
const U16_NAME: &str = type_name::<u16>();
const U32_NAME: &str = type_name::<u32>();
const U64_NAME: &str = type_name::<u64>();
const U128_NAME: &str = type_name::<u128>();

impl NumCast for Decimal {
    fn from<T: ToPrimitive>(n: T) -> Option<Self> {
        match type_name::<T>() {

            I8_NAME => return Decimal::from_i8(n.to_i8().unwrap()),
            I16_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            I64_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            I128_NAME => return Decimal::from_u8(n.to_u8().unwrap()),

            F32_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            F64_NAME => return Decimal::from_u8(n.to_u8().unwrap()),

            U8_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            U16_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            U32_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            U64_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            U128_NAME => return Decimal::from_u8(n.to_u8().unwrap()),
            _ => None
        }
    }
}
`

gechelberger:
Closes #573.

Fills a handful of `num_traits` gaps where `Decimal` already had the
underlying constants/methods but no trait impl, which forces downstream
generic code to special-case `Decimal`.

## Trait impls added

| Trait | Maps to | Location |
|---|---|---|
| `Bounded` | `Decimal::MIN` / `Decimal::MAX` | `src/decimal.rs` |
| `ConstZero` | `Decimal::ZERO` (const) | `src/decimal.rs` |
| `ConstOne` | `Decimal::ONE` (const) | `src/decimal.rs` |
| `NumCast` | `ToPrimitive::to_f64` → `Decimal::from_f64` | `src/decimal.rs` |
| `SaturatingAdd` | wraps inherent `Decimal::saturating_add` | `src/arithmetic_impls.rs` |
| `SaturatingSub` | wraps inherent `Decimal::saturating_sub` | same |
| `SaturatingMul` | wraps inherent `Decimal::saturating_mul` | same |

Also exposes `Bounded` through `rust_decimal::prelude` next to the existing
`One` / `Zero` / `Signed` re-exports. (`NumCast`/`Const*`/`Saturating*` are
typically used through generic bounds, not prelude imports — left out
intentionally.) 

This PR bumps the num_traits dependency from 0.2.0 to 0.2.18 in order to include ConstZero and ConstOne. My original plan was to leave it at 0.2.0 but after I had included the saturating operation traits I realized that they would force a bump to 0.2.12. I figured we could just include them all to see what the full changeset would look like. If you decide that you don't want to change the dependency resolution, it's trivial to strip those traits and trim the scope down to NumCast and Bounded.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -20,7 +20,7 @@ use diesel::{deserialize::FromSqlRow, expression::AsExpression, sql_types::Numer
 #[allow(unused_imports)] // It's not actually dead code below, but the compiler thinks it is.
 #[cfg(not(feature = "std"))]
 use num_traits::float::FloatCore;
-use num_traits::{FromPrimitive, Num, One, Signed, ToPrimitive, Zero};
+use num_traits::{Bounded, ConstOne, ConstZero, FromPrimitive, Num, One, Signed, ToPrimitive, Zero};
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
 #[cfg(all(target_arch = "wasm32", feature = "wasm"))]
@@ -2041,6 +2041,14 @@ impl One for Decimal {
     }
 }
 
+impl ConstZero for Decimal {
+    const ZERO: Self = Decimal::ZERO;
+}
+
+impl ConstOne for Decimal {
+    const ONE: Self = Decimal::ONE;
+}
+
 impl Signed for Decimal {
     fn abs(&self) -> Self {
         self.abs()
@@ -2080,6 +2088,22 @@ impl Num for Decimal {
 
     fn from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr> {
         Decimal::from_str_radix(str, radix)
+    }
+}
+
+impl Bounded for Decimal {
+    fn min_value() -> Self {
+        Decimal::MIN
+    }
+
+    fn max_value() -> Self {
+        Decimal::MAX
+    }
+}
+
+impl num_traits::NumCast for Decimal {
+    fn from<T: ToPrimitive>(n: T) -> Option<Self> {
+        n.to_f64().and_then(<Decimal as FromPrimitive>::from_f64)
     }
 }
 
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -2103,7 +2103,20 @@ impl Bounded for Decimal {
 
 impl num_traits::NumCast for Decimal {
     fn from<T: ToPrimitive>(n: T) -> Option<Self> {
-        n.to_f64().and_then(<Decimal as FromPrimitive>::from_f64)
+        // f64 can only exactly represent integers up to 2^53, and Decimal::MAX
+        // exceeds f64's integer range, so funneling everything through to_f64
+        // loses precision for large integer inputs. When the source is
+        // integer-valued, prefer the wide-integer path so values like
+        // 9_007_199_254_740_993u64 or i128 values near Decimal::MAX round-trip.
+        let f = n.to_f64()?;
+        if f.is_finite() && f.fract() == 0.0 {
+            if let Some(i) = n.to_i128() {
+                return Decimal::from_i128(i);
+            }
+            // we don't need to check n.to_u128() because no value in the interval
+            // (i128::MAX, u128::MAX] is representable by Decimal.
+        }
+        Decimal::from_f64(f)
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ wasm-bindgen = { default-features = false, version = "0.2", optional = true }
 [dev-dependencies]
 criterion = { default-features = false, version = "0.5" }
 csv = "1"
+proptest = { default-features = false, features = ["std"], version = "1.0" }
 rust_decimal_macros = { path = "macros" }
 serde = { default-features = false, features = ["derive"], version = "1.0" }
 serde_json = "1.0"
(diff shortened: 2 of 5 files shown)


src/decimal.rs
```diff
@@ -2083,6 +2091,22 @@ impl Num for Decimal {
...
+    }
+
+    fn max_value() -> Self {
+        Decimal::MAX
+    }
+}
+
+impl num_traits::NumCast for Decimal {
+    fn from<T: ToPrimitive>(n: T) -> Option<Self> {
+        n.to_f64().and_then(<Decimal as FromPrimitive>::from_f64)
```
paupino: The other contracts look good, but the implementation of this one I feel is problematic. For example:

```rust
        use num_traits::FromPrimitive;
        let n: u64 = 9_007_199_254_740_993;
        assert_eq!(
            <Decimal as n>::from(n),
            Some(Decimal::from_u64(n).expect("Parsed successfully"))
        );
```

Will fail with:
```
assertion `left == right` failed
  left: Some(9007199254740992)
 right: Some(9007199254740993)
```
This feels tricky to justify as a rounding error, given it's an integer.

Another one:
```rust
        use num_traits::FromPrimitive;
        let n: i128 = 79_228_162_514_264_337_593_543_950_000; // close to Decimal::MAX
        assert_eq!(
            <Decimal as NumCast>::from(n),
            Some(Decimal::from_i128(n).expect("Parsed successfully"))
        );
```

Will fail with:
```
assertion `left == right` failed
  left: None
 right: Some(79228162514264337593543950000)
```

My concern here is that you get subtle differences depending on how you parse the number, which I would consider as unexpected - given it's target is a `Decimal`. I think the latter example is validated somewhat in the docs for `NumCast`:

> If the source value cannot be represented by the target type, then None is returned.

I actually think this interface is a really tough one to work with and not really made for numbers outside of the `ToPrimitive` set. The only solution I can think of that kind of meets in the middle is to use `f64` but if there is no `fract` then fallback to wide integer arithmetic. It wouldn't be the "performant" path, but may reduce some confusion over conversion mistmatches.

e.g. (untested, but something like this)
```rust
  impl num_traits::NumCast for Decimal {
      fn from<T: ToPrimitive>(n: T) -> Option<Self> {
          let f = n.to_f64()?;
          if f.is_finite() && f.fract() == 0.0 {
              if let Some(i) = n.to_i128() { return Decimal::from_i128(i); }
              if let Some(u) = n.to_u128() { return Decimal::from_u128(u); }
          }
          Decimal::from_f64(f)
      }
  }
```

I think the other thing we need here is more test coverage just to make sure we're covering alternate angles.

paupino (changes requested):

src/decimal.rs
gechelberger: That's a good point. 

I think we can drop the second u128 fallback. It should always return None since the extrema of i128 are already outside the bounds of what Decimal can represent.

Either way, I think that this resolves your round-trip concern. I'll update it and include a test suite to prove to myself that it actually does what I expect.

gechelberger:
I took your proposed solution and it seems like it does what we want. The one exception is that f32 doesn't round trip exactly but instead takes on an f64 flavor. I don't think we can do better. Either we accept this or choose not to implement NumCast for Decimal (which seems like a perfectly defensible position to me).

Since proptest was already an optional feature on the cate, I added proptest as a dev-dependency to get proptest style coverage for the desired contracts

PR #804, opened 2026-06-12
fix: subtraction using carry-propogation can generate incorrect results

paupino:
Discovered a scenario where subtractions could provide incorrect results due to an incorrect break condition on carry. e.g.

```
34028236693 - 7.0000000000000000000000000000 = 68056473378.093846346337460743
34028236701 - 7.0000000000000000000000000000 = 1.906153653662539257
```

This fixes this issue by breaking when the value wraps, and not when it is zero.

This was a boundary condition case when the leftover was 1 or 0 (inside a u32, so rare - but definitely a hidden footgun).


paupino:
I think I'll hotfix this one.

PR #813, opened 2026-06-26
General clippy cleanup

paupino:
This is a general clippy cleanup:

1. Enable the `missing_const_for_fn` lint by default and add `const` where it is missing
2. Remove unused `InvalidDecimal` construct
3. Fixes a range clippy in `diesel` feature


Tony-Samuels (approved):
Seems fine, but while it does point out potential improvements, we do need to bear in mind that anything we make const now would be a breaking change if we wanted to add non-const usage to that function.

I think these particular changes are likely fine though.

PR #815, opened 2026-07-15
Fix round_sf returning an extra significant figure when rounding carries

gaoflow:
`round_sf` returns one more significant figure than requested when the rounding step carries into a new leading digit:

```rust
dec!(0.95).round_sf(1)  // 1.0, expected 1
dec!(9.95).round_sf(2)  // 10.0, expected 10
dec!(0.099).round_sf(1) // 0.10, expected 0.1
```

Trailing decimal zeros are significant here (the docs show `round_sf(305.459, 7) == 305.4590`), so these results carry an extra figure. It also makes `round_sf` non-idempotent: `round_sf(1.0, 1) == 1`.

The fractional branch rounds to a fixed number of decimal places, which assumes the integral digit count doesn't change; a carry through a run of nines adds one. The digit count only grows when the rounded result is exactly a power of ten, so the extra digit is always a trailing zero and dropping it is exact. Integral results (`9.95 -> 10`) are left as they are since the scale cannot go negative and their trailing zero isn't significant there.

The mantissa digit-count loop is extracted into a private helper so the carry check can reuse it. Tests cover the carry at several magnitudes and strategies, the legitimate trailing-zero cases, and a deterministic sweep asserting the figure count and idempotence for every strategy.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1611,22 +1611,6 @@ impl Decimal {
         self.round_sf_with_strategy(digits, RoundingStrategy::MidpointNearestEven)
     }
 
-    // Counts the digits making up the mantissa, ignoring sign and scale - effectively a naive
-    // log10. Zero has no significant figures.
-    fn significant_figures(&self) -> u32 {
-        let mut working = self.mantissa_array3();
-        let mut count = 0;
-        while !ops::array::is_all_zero(&working) {
-            let _remainder = ops::array::div_by_u32(&mut working, 10u32);
-            count += 1;
-            if working[2] == 0 && working[1] == 0 && working[0] == 1 {
-                count += 1;
-                break;
-            }
-        }
-        count
-    }
-
     /// Returns `Some(Decimal)` number rounded to the specified number of significant digits. If
     /// the resulting number is unable to be represented by the `Decimal` number then `None` will
     /// be returned.
@@ -1670,8 +1654,19 @@ impl Decimal {
             return Some(Decimal::ZERO);
         }
 
-        // We start by figuring out how many significant figures the mantissa is made up of.
-        let mantissa_sf = self.significant_figures();
+        // We start by grabbing the mantissa and figuring out how many significant figures it is
+        // made up of. We do this by just dividing by 10 and checking remainders - effectively
+        // we're performing a naive log10.
+        let mut working = self.mantissa_array3();
+        let mut mantissa_sf = 0;
+        while !ops::array::is_all_zero(&working) {
+            let _remainder = ops::array::div_by_u32(&mut working, 10u32);
+            mantissa_sf += 1;
+            if working[2] == 0 && working[1] == 0 && working[0] == 1 {
+                mantissa_sf += 1;
+                break;
+            }
+        }
         let scale = self.scale();
 
         match digits.cmp(&mantissa_sf) {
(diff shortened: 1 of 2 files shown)


src/decimal.rs
```diff
@@ -1727,7 +1732,17 @@ impl Decimal {
                     }
                     Some(num)
                 } else {
-                    Some(self.round_dp_with_strategy(scale - diff, strategy))
+                    let rounded = self.round_dp_with_strategy(scale - diff, strategy);
+                    // A round up that carries into a new leading digit (e.g. 0.95 -> 1.0) gains
+                    // an extra significant figure as a trailing zero; dropping it is exact.
+                    // Integral results (e.g. 9.95 -> 10) are left alone: the scale cannot go
+                    // negative and their trailing zero is not significant.
+                    let extra = rounded.significant_figures().saturating_sub(digits);
```
paupino: I'm a bit nervous about the second `significant_figures` calculation added here. It got me thinking, can this be achieved without this? I think it can... what about?

```rust
 if rounded.scale() > 0 && rounded.mantissa().unsigned_abs() == 10_u128.pow(digits) {
      Some(rounded.round_dp_with_strategy(rounded.scale() - 1, strategy))
  } else {
      Some(rounded)
  }
```

The reason this might work is because rounding away the low `diff` digits leaves `digits` figures, unless the round up carried into a new leading digit. The only way that happens is "9 + 1" - i.e. the mantissa is exactly 10^digits (e.g. 0.95 -> 1.0). Of course, the gained figure is a trailing zero, so dropping it is exact. Integral results (e.g. 9.95 -> 10) are left alone: the scale cannot go negative and their trailing zero is not significant.

The one thing I'd want to look at here still is the `10_u128.pow(digits)` call. We could work replace this with something like:

```rust
  #[inline(always)]
  const fn power_of_10(exp: u32) -> u128 {
      if (exp as usize) < POWERS_10.len() {
          POWERS_10[exp as usize] as u128
      } else if exp <= MAX_I64_SCALE {
          BIG_POWERS_10[exp as usize - 1] as u128
      } else {
          BIG_POWERS_10[MAX_I64_SCALE as usize - 1] as u128 * POWERS_10[(exp - MAX_I64_SCALE) as usize] as u128
      }
  }
```

We'd need to check the bench tests though - it may be a win, it may not.

paupino (commented):
Thanks for this - it's a good find! I added a comment as I think we can do this without another `div` loop (`div` is a slower operation than `mul`/`add`/`sub`) - but worth a sanity bench to confirm.

src/decimal.rs
binggao1230: Done — swapped the second `significant_figures()` pass for the carry check. You're right that the only way rounding adds a figure is the 9→10 carry, which lands the mantissa on exactly `10^digits`, so `scale() > 0 && mantissa == 10^digits` catches it and the `scale - 1` drop is exact. Full suite stays green, including the figure-count/idempotence sweep.

On `10_u128.pow(digits)`: I left it as the plain `pow` for now — in this branch `digits < mantissa_sf` (≤ 29), so it's bounded and already cheaper than the `div` loop it replaces. Happy to switch it to a `POWERS_10`/`BIG_POWERS_10` lookup if you'd prefer; I can run the round benches first to confirm it actually moves the needle before adding the helper.

src/decimal.rs
```diff
@@ -1654,19 +1670,8 @@ impl Decimal {
...
-        while !ops::array::is_all_zero(&working) {
-            let _remainder = ops::array::div_by_u32(&mut working, 10u32);
-            mantissa_sf += 1;
-            if working[2] == 0 && working[1] == 0 && working[0] == 1 {
-                mantissa_sf += 1;
-                break;
-            }
-        }
+        // We start by figuring out how many significant figures the mantissa is made up of.
+        let mantissa_sf = self.significant_figures();
```
paupino: Just as a bit of nit, it'd be good to unwind this now it's no longer required to be a function. It probably gets unwound during compilation anyway, but may as well make it explicit.

paupino (approved):

Issue #618, opened 2023-11-10
Rescaling a value with significant scale 28 to scale 29 creates a value with scale 29 and overflows on subtraction


david-monroe:
```rust
fn main() {
    let mut num = dec!(0.0000000000000000000000000001);
    dbg!(num);
    num.rescale(29);
    dbg!(num);
    dbg!(num.scale());
    dbg!(num - dec!(0.1));
}
```

```
[src/main.rs:6] num = 0.0000000000000000000000000001
[src/main.rs:8] num = 0.00000000000000000000000000010
[src/main.rs:9] num.scale() = 29
[src/main.rs:10] num - dec!(0.1) = -9999999999999999999999999990
```

Note that `num.rescale(30)` creates a value with scale 30 but does not overflow.

This happens when reading values with large scale from a database.

paupino:
Thanks for raising this. I performed some analysis and found that (long story short) scale 29 was causing an issue since it was "aligning" to a scale of 29 however reconstructing to a scale of 28. Because a scale of 29 is technically unsupported (except for very small numbers) it made sense to instead force scaling to 28 which effectively solves the issue.

To note, it didn't happen for scale of 30 since the very small number leaked into the high word triggering a full rescale. For scale 28 we didn't see the issue either since it was already within scale.

PR #816, opened 2026-07-17
Cap round_sf up-scaling at MAX_SCALE

gaoflow:
`round_sf` / `round_sf_with_strategy` up-scale the value when you request more significant figures than its magnitude holds. The target scale (`scale + digits - mantissa_sf`) wasn't bounded by `MAX_SCALE` (28), so it could return a `Decimal` with scale 29–31. Scale 31 panics on `Display`:

```rust
let d = Decimal::from_str("0.000000000000000000000000005").unwrap(); // 5e-27
println!("{}", d.round_sf(5).unwrap()); // panic: CapacityError at src/str.rs
```

Scale 29/30 don't panic but are still out of range (`0.05.round_sf(28)` → scale 29, `0.5.round_sf(29)` → scale 29). It's strategy-independent, since the up-scale runs before the strategy is applied. This is the up-scale sibling of #815's carry fix — same function, `Ordering::Greater` instead of `Ordering::Less`.

Fix: cap the target at `MAX_SCALE`. The value itself is representable (only the extra trailing zeros aren't), so the result is the value at max scale — `5e-27.round_sf(5)` → `0.0000000000000000000000000050`. That's already how `round_sf(u32::MAX)` behaves (`round_sf_large_digits_does_not_overflow`).

The doc says an unrepresentable result should be `None`, so returning `None` here is defensible too — but it reads the doc more strictly than current behavior: `1.5.round_sf(u32::MAX)` is `Some` today and `None` would regress that test. I went with the cap to keep the value; happy to switch to `None` if you'd rather.

Scope: checked the rest of the family against Python's `decimal` — `round_dp*` return early without up-scaling and `trunc*` only reduce scale, so neither can exceed `MAX_SCALE`. The shared `rescale_internal` can also produce scale 29–31 via the public `rescale()`, but that path looks intentional (issue #618 makes arithmetic on those values correct rather than clamping), so I left it untouched.

Test `round_sf_up_scale_stays_within_max_scale`: boundary/over-boundary inputs × all 7 strategies, asserting scale ≤ `MAX_SCALE` and that `to_string` doesn't panic, plus idempotence and in-range controls.

### DIFF
--- a/src/decimal.rs
+++ b/src/decimal.rs
@@ -1671,13 +1671,18 @@ impl Decimal {
 
         match digits.cmp(&mantissa_sf) {
             Ordering::Greater => {
-                // If we're requesting a higher number of significant figures, we rescale
+                // If we're requesting a higher number of significant figures, we rescale. The
+                // target is capped at MAX_SCALE: a Decimal cannot store a higher scale, and an
+                // uncapped value would be malformed (scale 29-31) and panic on Display.
                 let mut array = [self.lo, self.mid, self.hi];
                 let mut value_scale = scale;
                 ops::array::rescale_internal(
                     &mut array,
                     &mut value_scale,
-                    scale.saturating_add(digits).saturating_sub(mantissa_sf),
+                    scale
+                        .saturating_add(digits)
+                        .saturating_sub(mantissa_sf)
+                        .min(MAX_SCALE_U32),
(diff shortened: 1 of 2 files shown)


Tony-Samuels:
Thank you for the bug report & fix!

Issue #818, opened 2026-08-04
RUSTSEC-2026-0235 inactive rkyv 0.7 dependency triggers cargo audit


Phoenix500526:
`rust_decimal 1.42.1` declares `rkyv = "0.7.46"` as an optional dependency behind the `rkyv` feature.

The newly published [RUSTSEC-2026-0235](https://rustsec.org/advisories/RUSTSEC-2026-0235.html) affects the entire supported `rkyv 0.7` series:

> Insufficient archive validation can cause out-of-bounds reads in archives containing `Rc` or `Arc`.

The advisory is fixed in `rkyv >= 0.8.17`. The `0.7` series is affected and is no longer supported upstream.

# Minimal repro
```toml
# Cargo.toml
[dependencies]
 rust_decimal = "1.42.1"
```
```bash
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 1189 security advisories (from /Users/phoenix/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (73 crate dependencies)
Crate:     rkyv
Version:   0.7.46
Title:     Insufficient archive validation can cause out-of-bounds reads in archives containing Rc/Arc
Date:      2026-05-13
ID:        RUSTSEC-2026-0235
URL:       https://rustsec.org/advisories/RUSTSEC-2026-0235
Solution:  Upgrade to >=0.8.17

error: 1 vulnerability found!
```

paupino:
I'm tempted to break semver slightly and rip out `rkyv` v0.7. At the moment, we're keeping this feature for a small subset of users that are likely being pressured to upgrade `rkyv` nonetheless. This comes at the cost of inconveniencing the entire ecosystem (at least those using cargo audit/cargo deny).

It definitely goes against our last stance, hence the reasoning. I'll create a PR and let it simmer for a day or two to get feedback.

duaneking:
There is no legitimate reason to preserve backward compatibility with a dependency version that has known security vulnerabilities.

Backward compatibility is not a sufficient justification for retaining known-vulnerable dependencies. The affected version should be removed from supported configurations, and dependent versions should be upgraded to a secure dependency chain in accordance with standard vulnerability-management practices.

At this point, preserving the insecure version creates downstream security and dependency-management problems for everyone consuming the package. Consumers should not have to choose between compatibility and shipping software that passes their security gates.

[...]

PR #819, opened 2026-08-04
Remove rkyv 0.7 from feature bridge and update rand 0.8 support

### LINKED ISSUE #792: RUSTSEC-2026-0097: phantom `rand 0.8.5` in lockfile via optional `rand` feature
## Problem

`rust_decimal 1.41` declares `rand = "0.8"` as an optional dep (behind the `rand` feature). As of [RUSTSEC-2026-0097](https://rustsec.org/advisories/RUSTSEC-2026-0097.html) / [GHSA-cq8v-f236-94qc](https://github.com/advisories/GHSA-cq8v-f236-94qc) — "Rand is unsound with a custom logger using `rand::rng()`" (2026-04-09) — every `rand 0.8.x` version is flagged, and there is no patched 0.8.x (the fix landed in `0.9.3`).

Because cargo's resolver records optional-dep edges in `Cargo.lock` even when the activating feature is not enabled, every downstream crate that uses `rust_decimal` with `default-features = false, features = ["serde","std"]` (or similar minimal feature sets) still ends up with `rand 0.8.5` in its lockfile. Both `cargo audit --deny warnings` and GitHub Dependabot flag it. The dep is not actually compiled (verified below), but every downstream project still has to ship a local advisory ignore just to stay green.

## Minimal repro

```toml
# Cargo.toml
[dependencies]
rust_decimal = { version = "1.41", default-features = false, features = ["serde", "std"] }
```

```console
$ cargo audit
Crate:     rand
Version:   0.8.5
Title:     Rand is unsound with a custom logger using `rand::rng()`
Date:      2026-04-09
ID:        RUSTSEC-2026-0097
Dependency tree:
rand 0.8.5
└── rust_decimal 1.41.0
warning: 1 allowed warning found
```

Confirmed the dep is not actually pulled into the build:
- `cargo tree -e normal` shows no `rand` under `rust_decimal` (only `arrayvec`, `num-traits`, `serde`)
- verbose `cargo build --release` never invokes `rustc --crate-name rand`
- `nm libmycrate.so | grep -iE 'rand_core|rand_chacha|::rand::'` returns zero symbols

So the actual attack surface is zero — it's purely a lockfile/scanner noise problem, but it's the same pattern as #766 (RUSTSEC-2026-0001 via rkyv) and is going to keep surfacing until the 0.8-line `rand` dep is retired.

## Suggested fixes (any one works)

1. **Drop the `rand` (0.8) optional dep** in the next minor release, keeping only `rand-0_9` (added in #702). Document the migration for users who were enabling the `rand` feature.
2. **Alias the `rand` feature to `rand-0_9`** — i.e. `rand = ["rand-0_9"]` and remove `dep:rand`. Keeps the feature name stable while moving users to the patched line.
3. **Bump the version constraint** on the optional `rand` dep from `"0.8"` to `">=0.9.3"`. Breaking for anyone using rust_decimal's rand integration through the old API, but the rust_decimal-internal usage appears to be limited and migratable.

Happy to open a PR for any of these if one is preferred.

## Why this matters

Dependabot's low-severity alerts from transitively-carried optional deps accumulate quickly across a large user base, and every downstream project is independently writing a `.cargo/audit.toml` ignore + a dismissal comment. A single upstream drop would clear the advisory for everyone.

paupino:
Fixes two RUSTSEC advisories from downstream optional dependencies.
1. Removes `rkyv` 0.7 support (somewhat controversially) to address #818  
2. Updates `rand` dependency to 0.8.6 to fix #792 

My reasoning:
> I'm tempted to break semver slightly and rip out rkyv v0.7. At the moment, we're keeping this feature for a small subset of users that are likely being pressured to upgrade rkyv nonetheless. This comes at the cost of inconveniencing the entire ecosystem (at least those using cargo audit/cargo deny).

I'm open to feedback on this approach however since it is somewhat controversial. It comes from a position where we're dependent on downstream libraries maintaining legacy versions of their product.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,9 @@ ndarray = { default-features = false, optional = true, version = "0.15.6" }
 num-traits = { default-features = false, features = ["i128"], version = "0.2" }
 postgres-types = { default-features = false, optional = true, version = "0.2" }
 proptest = { default-features = false, optional = true, features = ["std"], version = "1.0" }
-rand = { default-features = false, optional = true, version = "0.8" }
+rand = { default-features = false, optional = true, version = "0.8.6" }
 rand-0_9 = { default-features = false, optional = true, package = "rand", version = "0.9" }
 rust_decimal_macros = { path = "macros", default-features = false, optional = true, version = "1.37.0" }
-rkyv = { default-features = false, features = ["size_32", "std"], optional = true, version = "0.7.46" }
 rocket = { default-features = false, optional = true, version = "0.5.0-rc.3" }
 zmij = { default-features = false, optional = true, version = "1.0" }
 serde = { default-features = false, optional = true, version = "1.0" }
@@ -62,7 +61,7 @@ bytes = { default-features = false, version = "1.0" }
 diesel = { default-features = false, features = ["mysql", "postgres"], version = "2.2.3" }
 futures = { default-features = false, version = "0.3" }
 postgres = { default-features = false, version = "0.19" }
-rand = { default-features = false, features = ["getrandom"], version = "0.8" }
+rand = { default-features = false, features = ["getrandom"], version = "0.8.6" }
 rust_decimal = { path = ".", features = ["macros"] }
 rand-0_9 = { default-features = false, features = ["thread_rng"], package = "rand", version = "0.9" }
 rkyv-0_8 = { version = "0.8.13", package = "rkyv" }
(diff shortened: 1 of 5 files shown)


Tony-Samuels:
How close are you to your goals for v2? The other option here is to cut our own breaking release.

paupino:
That's true. We could do away with the v1 and cut v2 - it already contains a few breaking changes. I'd need to do a quick review as once we commit to the API shape we're stuck with it (so want to make sure we're removing some of the footguns present in v1).

paupino:
I think since v2 is likely still very much pre-release (and may take a little bit to stabalise) I'm going merge this one. `cargo audit` is something that is common in build chains and this will help unblock them without having to create an exception for the explicit security notice. The consequence of merging is that it IS a breaking change for those still on `rkyv` of which I apologise. For those that break, the workaround is to either:
1. Pin at `1.42.1` or,
2. Upgrade `rkyv`

Overall, not ideal and something that I _really_ want to move away from in v2 (i.e. being flagged for dependencies that aren't even real dependencies).

PR #820, opened 2026-08-04
Removing v1 deadweight from v2 to avoid carrying it forward

paupino:
Tried removing a bunch of deadweight from v2 to avoid carrying it over. 
Includes trying to clean up serde a bit, which has traditionally been a complicated part of the library to understand.

Open questions:
1. Are the serde feature names clearer? Or does it make it worse? 
2. Thoughts on removing `new` and making it explicit? (Sorry, this is most of the change in this PR)
3. Are there any other features to drop and/or move to a new crate?


src/serde.rs
```diff
@@ -277,38 +273,32 @@ pub mod str_option {
...
-    fn deserialize<D>(deserializer: D) -> Result<Decimal, D::Error>
-    where
-        D: serde::de::Deserializer<'de>,
-    {
-        deserializer.deserialize_f64(DecimalVisitor)
+        // Self-describing formats (JSON, YAML) can accept either a string or a
+        // number, so `deserialize_any` is both safe and maximally permissive.
+        // Non-self-describing formats (bincode, postcard) reject `deserialize_any`
+        // outright and must be told the concrete type up front.
+        if deserializer.is_human_readable() {
```
paupino, line 285: This is new by the way - I discovered it in the docs and seems to be better than the gated approach previously

src/serde.rs
```diff
@@ -71,7 +71,7 @@ pub mod arbitrary_precision {
 ///     r#"{"value":null}"#
 /// );
 /// ```
-#[cfg(feature = "serde-with-arbitrary-precision")]
+#[cfg(feature = "serde-default-exact")]
 pub mod arbitrary_precision_option {
```
Tony-Samuels, line 75: I think we should keep the existing name here. It matches serde-json, which is the only deserializer that makes use of the behaviour.

README.md
Tony-Samuels, line 1: I think we should add a guarantee in the features section or similar on how we treat old versions of crates, e.g. rand.

My proposal would be something like:
```
We will endeavour to always support the most recent versions of crates. We will additionally support old versions until one of the following is true (extended support may be provided at the maintainers' choice):
- 3 months after the next version's release
- Security updates are no longer provided
- The version, or crate as a whole, is no longer maintained
```

Feels like that would solve 80 odd % of the breaking changes we'd need to make.

Tony-Samuels (commented):
I'll want to play around with the serde stuff before confirming. Just a couple of small things in the meantime.

src/serde.rs
paupino: Yeah, fair call. Let me take another stab at naming.

README.md
paupino: Yeah, this would help. 

Personally, I like the remote derive pattern that some libraries have picked up and would like to see that more widely used. It's always been a bit of a chicken/egg thing about where the code lives. I may do some exploring around this too - I think the _ideal_ state is a very limited set of dependencies that are effectively "std" and the rest can be bridged.

src/serde.rs
paupino: Renamed it back. I kept the default prefix, but only because I'm trying to convey "this overrides default behavior". Happy to revert that too, if unnecessary.

src/serde.rs
Tony-Samuels, line 350: This doesn't work anymore as `Decimal::from_str` falls back to `Decimal::from_scientific_lossy` internally.

src/serde.rs
```diff
@@ -369,7 +359,7 @@ impl<'de> serde::de::Visitor<'de> for DecimalVisitor {
             .map_err(|_| E::invalid_value(Unexpected::Str(value), &self))
     }
 
-    #[cfg(feature = "serde-with-arbitrary-precision")]
+    #[cfg(feature = "serde-default-arbitrary-precision")]
```
Tony-Samuels, line 362: Is there ever a case where having this as a thing we can handle is a bad thing? Why not just always include it?

src/serde.rs
Tony-Samuels: This applies to a few different places.

Tony-Samuels, line 93: Should this delegate to `super::arbitrary_precision::serialize`?

src/serde.rs
```diff
@@ -510,12 +488,12 @@ impl<'de> serde::de::Deserialize<'de> for DecimalKey {
     }
 }
 
-#[cfg(feature = "serde-with-arbitrary-precision")]
+#[cfg(feature = "serde-default-arbitrary-precision")]
```
Tony-Samuels, line 491: Thoughts on an arbitrary precision module with all of these in it? There's quite a few types where the only indication that they're arbitrary precision specific is their feature requirement.

Would also mean we'd need the `cfg` in far fewer places.

src/serde.rs
```diff
@@ -545,7 +523,7 @@ impl<'de> serde::de::Deserialize<'de> for DecimalFromString {
     }
 }
 
-#[cfg(not(feature = "serde-float"))]
+#[cfg(not(feature = "serde-default-number"))]
```
Tony-Samuels, line 526: Could consider using https://docs.rs/cfg-if/latest/cfg_if/ for these, so that we can make it a bit easier to write. Something like:
```rust
cfg_if::cfg_if! {
    if #[cfg(all(feature = "serde-default-number", feature = "serde-default-arbitrary-precision"))] {
        // `serde_json::Number` is JSON-specific: a binary format receives its
        // internal representation rather than a number. Fall back to the plain
        // numeric wire form there, matching `serde-default-number` alone.
        if !serializer.is_human_readable() {
            use num_traits::ToPrimitive;
            return serializer.serialize_f64(self.to_f64().unwrap());
        }
            .serialize(serializer)
    } else if #[cfg(feature = "serde-default-number")] {
        use num_traits::ToPrimitive;
        serializer.serialize_f64(self.to_f64().unwrap())
    } else {
        let value = crate::str::to_str_internal(self, true, None);
        serializer.serialize_str(value.0.as_ref())
    }
}
```

Main advantage is not needing all the `not(feature)` handling, but putting it inside the `fn serialize` block also makes it easier to see all the impls at a glance, at least when they're all small like here.

src/serde.rs
```diff
@@ -71,7 +71,7 @@ pub mod arbitrary_precision {
 ///     r#"{"value":null}"#
 /// );
 /// ```
-#[cfg(feature = "serde-with-arbitrary-precision")]
+#[cfg(feature = "serde-default-arbitrary-precision")]
```
Tony-Samuels, line 74: It may be this is explicitly how you want to handle it, but having this only supported behind `default` means that if you want one of your fields to be arbitrary precision, every other Decimal being de/serialized now is too, even in other crates implementations.

PR #827, opened 2026-09-20
Fix square-root iteration cycles

### LINKED ISSUE #721: Panic in sqrt function
```use rust_decimal::MathematicalOps;
use rust_decimal_macros::dec;

fn main() {
    let a = dec!(400.00000000000000000000000003);
    a.sqrt();
}
```

Will panic here https://github.com/paupino/rust-decimal/blob/63f5ca25a8712a1c78bb10f1e1aefd30707b81d7/src/maths.rs#L359 

```result: 20.00000000000000000000 last: 20.000000000000000000000000001
result: 20.000000000000000000000000001 last: 20.00000000000000000000
result: 20.00000000000000000000 last: 20.000000000000000000000000001
result: 20.000000000000000000000000001 last: 20.00000000000000000000
result: 20.00000000000000000000 last: 20.000000000000000000000000001
```

RanaPriyansh:
Fixes #721.

For some positive inputs, `sqrt` alternates between two Decimal estimates until the circuit breaker panics. Detect a repeated estimate with the same value and scale, then return the larger endpoint. Keep the existing fixed-point check, arithmetic update and iteration limit.

Adds regression coverage for the reported input and preserves explicit negative-zero behavior. The change does not promise correctly rounded results for every input or address cycles longer than two.

Validation: focused regression and controls; workspace maths and maths-nopanic tests; default-feature maths tests; formatting; all-feature workspace Clippy; and the library check without default features. A private comparison preserved 861 previously convergent results and ten domain results; all 15 measured two-value cycles terminated at the independently checked upper endpoint.

Implemented and checked with OpenAI Codex assistance. Cross-platform CI and maintainer review remain pending.


Tony-Samuels:
This seems plausible, but I don't know if cycles of length-2 is the right coverage here. Looping in @paupino who knows a lot more of the decimal maths.

